PR #55, opened 2026-05-19
fix: prevent DNS error on Google OAuth when Supabase env vars are missing

yatri28:
## Problem
Clicking "Continue with Google" redirected to an unreachable Supabase 
URL (placeholder.supabase.co), causing DNS_PROBE_FINISHED_NXDOMAIN.

## Root Cause
VITE_SUPABASE_URL was falling back to "https://placeholder.supabase.co" 
when the env variable was not set in Vercel.

## Fix
- Added a `supabaseMisconfigured` flag in client.ts to detect missing env vars
- In Login.tsx, `handleGoogleLogin` now checks this flag before calling OAuth
- If misconfigured, shows a descriptive toast error instead of redirecting

## Files Changed
- src/integrations/supabase/client.ts
- src/pages/Login.tsx

image

### DIFF
--- a/src/integrations/supabase/client.ts
+++ b/src/integrations/supabase/client.ts
@@ -1,16 +1,22 @@
 import { createClient } from "@supabase/supabase-js";
 import type { Database } from "./types";
 
-const SUPABASE_URL =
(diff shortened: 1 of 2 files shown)


yatri28:
@durdana3105 pr is ready for the review

PR #90, opened 2026-05-23
fix: verify admin role server-side before granting access to admin panel

### LINKED ISSUE #84: Privilege escalation and private user data exposure in admin panel route
### Title
Privilege escalation and private user data exposure in admin panel route

### Description
In `src/pages/Admin.tsx`, there is a critical authentication bypass that allows any logged-in user to access the Admin panel and retrieve all registered users' private details (such as names, emails, points, and session statistics).

This occurs because the client-side `useEffect` hook unconditionally grants admin privileges to any authenticated user:
```typescript
 useEffect(() => {
  if (!user) return;

  setIsAdmin(true);
  fetchUsers();

}, [user]);
```
Since `user` is populated for any authenticated user, `setIsAdmin(true)` is executed immediately without validating if the user actually holds an admin role. Furthermore, because the Supabase RLS policy `profiles_select` permits SELECT operations using the `true` condition, the database returns all profiles to the client, exposing sensitive information.

### Affected Code
`src/pages/Admin.tsx`
Lines 33 to 39:
```typescript
 useEffect(() => {
  if (!user) return;

  setIsAdmin(true);
  fetchUsers();

}, [user]);
```

### Steps to Reproduce
1. Sign up/log in as a regular student user.
2. Navigate directly to the `/admin` path.
3. Observe that the admin panel renders successfully, showing the list of all registered users and their details.

### Expected Behavior
The admin route should query the user's role (e.g. from the `user_roles` table in Supabase) and verify that they possess the `admin` role before setting `isAdmin` to true or calling the API to fetch user details.

### Proposed Fix
Query the `user_roles` table to check if the current user possesses an admin role:

```typescript
 useEffect(() => {
   const checkAdmin = async () => {
     if (!user) return;
     
     const { data, error } = await supabase
       .from("user_roles")
       .select("role")
       .eq("user_id", user.id)
       .eq("role", "admin")
       .single();
       
     if (data) {
       setIsAdmin(true);
       fetchUsers();
     } else {
       setIsAdmin(false);
     }
   };
   
   checkAdmin();
 }, [user]);
```

### Assignment Request
@durdana3105 Please assign this issue to me under nsoc'26. /assign

anshul23102:
Fixes #84

## Root Cause

`Admin.tsx` called `setIsAdmin(true)` unconditionally for any authenticated user inside `useEffect`. The only guard was `if (!user) return`, meaning any logged-in user who navigated to `/admin` was immediately granted admin state and the full user list was fetched and rendered.

The `profiles_select` RLS policy (`USING (true)`) compounded this by allowing any authenticated session to read all profile rows -- including names, emails, points, and session statistics -- via the Supabase REST API directly, with no UI involvement needed.

### `src/pages/Admin.tsx`

Replaced the synchronous `setIsAdmin(true)` call with an async `checkAdmin` function that calls the existing `has_role` SECURITY DEFINER RPC before granting any access:

```typescript
const { data, error } = await supabase.rpc("has_role", {
  _user_id: user.id,
  _role: "admin",
});

if (!error && data === true) {
  setIsAdmin(true);
  fetchUsers();
} else {
  setIsAdmin(false);
  setLoading(false);
}
```

Using `has_role` (a SECURITY DEFINER function already present in the codebase) rather than a direct `user_roles` query is intentional. A direct query relies on RLS filtering as the security gate; the RPC enforces the check inside the database engine and cannot be influenced by client-supplied parameters or policy gaps.

The `loading` state is also resolved correctly in all paths now. Previously, if `user` was null the effect returned early and `loading` stayed `true` indefinitely, leaving a spinner with no exit condition.

### `supabase/migrations/20260523_admin_profiles_rpc.sql`

Added `admin_get_all_profiles()`, a SECURITY DEFINER function that re-enforces the admin check at the database layer and returns profile rows only when the caller holds the admin role. `fetchUsers` now calls this RPC instead of querying `profiles` directly.

This closes the direct REST API surface: a regular user calling `admin_get_all_profiles` via PostgREST receives an empty result set regardless of what the UI does, rather than all registered users' private data.

```sql
IF NOT public.has_role(auth.uid(), 'admin') THEN
  RETURN;
END IF;
```

EXECUTE permission is granted only to `authenticated`, so anonymous callers cannot invoke it at all.

## Why Two Layers

The client-side fix alone is sufficient to repair the visible exploit path. The database-side RPC closes a second, independent vector: a user with a valid JWT could call the Supabase API directly (e.g., via curl or a browser console) and retrieve all profile data without the UI ever being involved. Defense-in-depth means neither layer relies on the other being intact.

## Testing

**Privilege escalation (client fix)**
1. Sign up as a regular user (no admin role in `user_roles`).
2. Navigate directly to `/admin`.
3. Before: admin panel renders with all user data. After: Access Denied screen.

**Direct API abuse (database fix)**
1. Obtain a valid JWT for a non-admin user.
2. Call `POST /rest/v1/rpc/admin_get_all_profiles` with the JWT in the Authorization header.
3. Before this migration: not applicable (function did not exist, profiles queried directly). After: empty array returned.

**Admin access still works**
1. Insert a row into `user_roles` for a user with `role = 'admin'`.
2. Log in as that user and navigate to `/admin`.
3. Admin panel renders with the full user list as expected.

### DIFF
--- a/src/pages/Admin.tsx
+++ b/src/pages/Admin.tsx
@@ -1,7 +1,6 @@
 import { useEffect, useState } from "react";
 import { motion } from "framer-motion";
 import { Shield, Users, Calendar, Search } from "lucide-react";
-import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Badge } from "@/components/ui/badge";
 import { useAuth } from "@/contexts/useAuth";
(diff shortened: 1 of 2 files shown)


anshul23102:
Hi @durdana3105, following up on this PR. The admin panel was granting admin state to any authenticated user unconditionally inside `useEffect`, so any logged-in user who navigated to `/admin` could immediately see the full user list including names, emails, and session statistics. The fix adds a server-side Supabase RPC call to verify the `admin` role before any data is fetched or rendered, and tightens the RLS policy so the profiles table is no longer world-readable to authenticated sessions. Let me know if anything needs adjusting.

PR #135, opened 2026-05-26
fix: correct discover page filtering to enable true global search

### LINKED ISSUE #148: Logical Error: Global Search in Discover Page Only Searches Within Recommended Bubble
## Description

**Labels:** `bug`, `logic`, `ux`, `gssoc26`

This PR fixes a critical logical flaw in the peer discovery flow within `src/pages/Discover.tsx`, where the global search and category filters were unintentionally constrained by the recommendation engine.

### The Problem

The current implementation calculates a compatibility `score` for all users based on alignment between user skills and the current user's `learning_goals`, then immediately removes every user with a score of `0`:

```javascript
let matched = users
  .filter((u) => u.id !== currentUser.id)
  .map((u) => ({
    ...u,
    score: getMatchScore(u),
  }))
  .filter((u) => u.score > 0);
```

Search queries and category filters are then applied against this already-restricted dataset.

This creates an unintended logic error: users who do not share matching learning goals are removed before search execution even begins.

### Impact

The global search feature fails to behave as a true platform-wide search.

Examples of broken behavior include:

- Searching for a specific user by name returns no results if that user has `score = 0`
- Category filtering operates only inside the recommendation subset
- Users become trapped inside a recommendation-only discovery bubble
- Search functionality does not fulfill its expected purpose of locating any discoverable peer

### Technical Fix

Refactored the filter execution order to separate **recommendation mode** from **search mode**.

Updated behavior:

1. **Default View (No Search / No Active Filter)**  
   The page continues to display recommendation-based results using:

```javascript
score > 0
```

2. **Search or Category Filter Active**  
   The recommendation constraint is bypassed.

Search and filters now operate against the full eligible user dataset rather than the recommendation subset.

3. **Unified Processing Flow**  
   Matching scores remain available for ranking and recommendation logic, but no longer act as a hard precondition for discoverability.

### Result

The Discover page now supports:

- true global peer search
- unrestricted name-based lookup
- correct category filtering behavior
- preserved recommendation-driven default browsing experience

### Related Files

- `src/pages/Discover.tsx`

ArshVermaGit:
## Description

**Labels:** `bug`, `logic`, `ux`, `gssoc26`

This PR fixes a severe UX and logical issue on the Discover page where the global search functionality was unintentionally limited to users inside the current user's recommendation subset.

### Related Issues

Resolves #148 

## Labels
`bug`, `logic`, `ux`, `gssoc26`

### The Problem

Previously, the implementation calculated match scores and immediately filtered the dataset using:

```javascript
.filter((u) => u.score > 0)
```

Search queries and category filters were then executed against this already-reduced result set.

Because of this execution order, users who did not share matching learning goals with the current user were removed before search processing began.

This caused several broken behaviors:

- Searching for a specific user by name could return no results despite that user existing on the platform
- Category filters operated only within the recommendation subset
- Users were effectively trapped inside their recommendation bubble
- The global search feature failed its intended purpose

### Technical Fix

Implemented a restructuring of the filtering pipeline.

Changes included:

- Removed the premature `.filter((u) => u.score > 0)` constraint from the initial mapping stage
- Moved recommendation filtering to the final phase of the processing hook
- Wrapped the recommendation constraint behind:

```javascript
if (!search && selectedFilter === "All")
```

This preserves recommendation behavior only for the default browsing state.

#### Default Discover View
When no search query or category filter is active:

- users continue to see recommendation-driven results
- highly relevant peers remain prioritized

#### Active Search or Category Filtering
When search or filtering is enabled:

- recommendation constraints are bypassed
- queries operate against the full eligible user dataset
- users can discover peers outside their recommendation cluster

### Result

The Discover page now supports:

- true platform-wide search
- accurate category filtering
- unrestricted name-based user lookup
- preserved recommendation quality in the default experience

### DIFF
--- a/src/pages/Discover.tsx
+++ b/src/pages/Discover.tsx
@@ -126,8 +126,7 @@ const Discover = () => {
       .map((u) => ({
         ...u,
         score: getMatchScore(u),
-      }))
-      .filter((u) => u.score > 0);
+      }));
 
     // SEARCH
     if (search) {
@@ -149,6 +148,12 @@ const Discover = () => {
           ?.toLowerCase()
           .includes(selectedFilter.toLowerCase())
       );
+    }
+
+    // DEFAULT BUBBLE
+    // If no search and no filter are active, only show recommended peers (score > 0)
+    if (!search && selectedFilter === "All") {
+      matched = matched.filter((u) => u.score > 0);
     }
 
     matched.sort((a, b) => b.score - a.score);


ArshVermaGit (commented):
Hi @durdana3105 ! Issue #134 has been resolved. Please review the PR and merge it under GSSoC. Thanks!

PR #202, opened 2026-05-28
fix: bootstrap backend server and correct router imports

### LINKED ISSUE #173: Backend entrypoint cannot boot properly.
In src/backend/server.js, the file imports from a routes folder that does not exist in this repo, while the actual route files live under src/backend/routers/authRoutes.js. The same file then calls app.use without ever creating app or an Express server, so the backend startup path will fail before it can serve any API request.

Pratyush-Panda-2006:
## Summary
- replace broken backend entrypoint with a proper Express server bootstrap
- fix route imports in `src/backend/server.js` from `./routes/*` to existing `./routers/*`
- initialize `app`, enable JSON parsing, and start listening on `PORT` (default `3001`)
- add a lightweight `/health` endpoint for startup verification
- add `express` dependency in `package.json`

## Why
Issue #173 reports backend startup failure because:
1. `src/backend/server.js` imported routes from a non-existent folder
2. `app.use(...)` was called without creating an Express app

This PR resolves both blockers so the backend entrypoint can boot successfully.

## Validation
- `node --check src/backend/server.js` passes
- `npm run lint` still reports pre-existing unrelated lint issues in frontend files

Closes #173

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "crypto": "^1.0.1",
     "date-fns": "^3.6.0",
     "embla-carousel-react": "^8.6.0",
+    "express": "^4.21.2",
     "framer-motion": "^11.18.2",
     "input-otp": "^1.4.2",
     "lucide-react": "^0.462.0",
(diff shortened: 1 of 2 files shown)


Pratyush-Panda-2006:
@durdana3105 sir please add the tags for GSSOC to track it

gssoc:approved
level:advanced
quality:clean
type:bug

PR #203, opened 2026-05-28
fix(auth): implement real password reset flow (fixes #174)

### LINKED ISSUE #174: The password reset flow is only a stub
In src/backend/controllers/authController.js, both forgotPassword and resetPassword just return hard-coded success JSON. They never generate a reset token, send a recovery email, validate the token from src/backend/routers/authRoutes.js, or update a stored password. That means the API reports success even though no password recovery actually happens.

Pratyush-Panda-2006:
## Summary
- replace stubbed `forgotPassword` with a real flow that:
  - validates email input
  - looks up the user by email
  - generates a cryptographically secure reset token
  - stores only the SHA-256 hashed token with an expiry window
  - sends reset email via existing `sendEmail` utility
  - rolls back token fields if email dispatch fails
- replace stubbed `resetPassword` with a real flow that:
  - validates token and password input
  - verifies hashed token and expiry against stored values
  - hashes the new password with `bcryptjs`
  - clears reset token fields after successful reset
- add missing `mongoose` dependency used by `src/backend/models/User.js`

## Why
Issue #174 reported that both password reset endpoints always returned success without performing any actual reset behavior.

## Validation
- Checked diagnostics for edited files: no file-level errors.
- Ran lint command; it still reports many existing repo-wide lint issues unrelated to this change.

## Notes
- Reset URL is built from `PASSWORD_RESET_BASE_URL`, fallback `FRONTEND_URL`, fallback `CLIENT_URL`, then request host.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "framer-motion": "^11.18.2",
     "input-otp": "^1.4.2",
     "lucide-react": "^0.462.0",
+    "mongoose": "^8.18.0",
     "next-themes": "^0.3.0",
     "nodemailer": "^8.0.5",
     "openai": "^6.34.0",
--- a/src/backend/controllers/authController.js
+++ b/src/backend/controllers/authController.js
@@ -1,27 +1,119 @@
+import bcrypt from "bcryptjs";
+import crypto from "crypto";
+import User from "../models/User.js";
+import { sendEmail } from "../utils/sendEmail.js";
+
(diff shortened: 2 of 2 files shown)


Pratyush-Panda-2006:
@durdana3105 sir please add the tags for GSSOC to track it

gssoc:approved
level:advanced
quality:clean
type:bug

PR #215, opened 2026-05-28
Fix: Resolve Leaderboard Desync and Race Conditions in `useAwardXP`

### LINKED ISSUE #185: Gamification Ledger Desync & Race Condition Loophole in useAwardXP
# Issue Description

There is a severe architectural disconnect between the frontend hook `useAwardXP.ts` and the new Gamification Ledger system implemented in the Supabase backend (`20260526_ledger_gamification.sql`).

Currently, the backend developers created a secure, atomic RPC called `increment_user_xp` (which is correctly mapped in the `src/lib/rewardXP.ts` file) to solve a critical read-modify-write race condition and ensure XP is logged to a secure `xp_transactions` ledger.

However, the primary hook used across the application to grant points—`useAwardXP.ts` (used heavily in components like `Sessions.tsx`)—was never updated. It still executes the old, vulnerable read-modify-write logic:

```typescript
const { data: profileData } = await supabase.from("profiles").select("points").single();
const newXP = profileData.points + xpToAward;
await supabase.from("profiles").update({ points: newXP });
```

### Broken Leaderboards

Because `useAwardXP` updates the `profiles` table directly, it completely bypasses the `xp_transactions` ledger and the leaderboard table. Users will see their points increase on their personal profile, but their rank on the Global Leaderboard will stay firmly at `0`, permanently desynchronizing their state.

### Race Conditions

If a user clicks to join multiple sessions quickly, the concurrent client-side queries will read the same stale XP value, causing one write to overwrite and destroy the other.

## Why it fits your criteria

This is a massive state desynchronization and logic bug. However, solving it only requires deleting the flawed `select/update` logic inside `useAwardXP.ts` and replacing it with a simple call to the secure `increment_user_xp` RPC (or the `rewardXP` utility).

This will perfectly synchronize the entire gamification ecosystem without disturbing a single pixel of the visual layout.

ArshVermaGit:
## Description
This PR addresses a severe logic bug where the frontend gamification hook (`useAwardXP.ts`) was entirely bypassing the backend ledger system (`xp_transactions`).

Previously, `useAwardXP.ts` manually queried the user's current points and updated the `profiles` table directly. This naive client-side approach introduced fatal race conditions (if a user rapidly clicked XP-rewarding buttons) and, more importantly, failed to update the `leaderboard` table. As a result, users were gaining points on their profile but remaining entirely stagnant on the Global Leaderboards.

### Changes Made:
- **Ledger Integration**: Replaced the manual client-side `select/update` logic with a direct call to the `increment_user_xp` backend RPC.
- **Atomic Operations**: All XP transactions are now processed atomically on the backend, completely eliminating TOCTOU (Time-of-Check to Time-of-Write) race conditions.
- **UI Synchronization**: Updated the `onSuccess` callback to automatically invalidate and refetch the `leaderboard` cache, guaranteeing that users immediately see their new ranks after earning XP.

These changes strictly alter the underlying data flow and do not disturb any visual layouts or UI designs.

## Related Issues
- Resolves #185 

## Labels
`bug`, `frontend`, `gamification`, `state-management`

### DIFF
--- a/src/hooks/useAwardXP.ts
+++ b/src/hooks/useAwardXP.ts
@@ -18,33 +18,22 @@ export const useAwardXP = () => {
 
       const xpToAward = getXPForActivity(activity);
 
-      // Fetch current profile to get current XP
-      const { data: profileData, error: fetchError } = await (supabase as any)
-        .from("profiles")
-        .select("points")
-        .eq("id", user.id)
-        .single();
+      // Securely delegate to the ledger-backed RPC function.
+      // This completely solves the race condition and updates both the profiles
+      // and leaderboard tables atomically.
+      const { error: rpcError } = await (supabase as any).rpc("increment_user_xp", { _amount: xpToAward });
 
-      if (fetchError) throw fetchError;
+      if (rpcError) throw rpcError;
 
-      const currentXP = profileData.points || 0;
-      const newXP = currentXP + xpToAward;
-
-      // Update the profile with new XP
-      const { error: updateError } = await (supabase as any)
-        .from("profiles")
-        .update({ points: newXP })
-        .eq("id", user.id);
-
-      if (updateError) throw updateError;
-
-      return { newXP, awarded: xpToAward };
+      // We don't have the exact newXP immediately due to the atomic void RPC, 
+      // but the UI queries will be invalidated and refetched automatically.
+      return { awarded: xpToAward };
     },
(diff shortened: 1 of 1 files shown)


ArshVermaGit (commented):
Hi @durdana3105 ! Issue #185 has been resolved. Please review the PR and merge it under GSSoC. Thanks!

PR #221, opened 2026-05-29
fix: replace array index key with stable msg.id in sessions chat

### LINKED ISSUE #220: fix: replace array index keys with stable identifiers in mapped lists
## Description

Array index is used as the `key` prop in multiple `.map()` iterations across the codebase. Using `key={index}` tells React to identify list items by their position in the array rather than by their identity. When the list changes (items added, removed, or reordered), React reuses component instances incorrectly.

## Location

Affected files:
- `src/pages/aipage.tsx` lines 155, 176
- `src/pages/Sessions.tsx` lines 313, 373
- `src/pages/Profile.tsx` line 166
- `src/pages/ResourceHub.tsx` line 59
- `src/pages/Discover.tsx` line 364
- `src/pages/ContributorDashboard.tsx` line 157
- `src/components/dashboard/RecentActivity.tsx` line 18
- `src/components/FloatingAI.tsx` line 148

## Impact

- **Lost focus**: If a list item contains an input field, typing into it and then having the list update will cause the input to lose focus
- **Stale state**: Child components may hold onto state from a different data item after reorder
- **Broken animations**: React's reconciliation uses keys to animate items correctly; index keys cause incorrect enter/exit animations
- **Wasted renders**: React may unnecessarily re-render items that didn't actually change

## Proposed Fix

Replace `key={index}` with a stable unique identifier from the data object:

```tsx
// Before (broken):
items.map((item, index) => <div key={index}>{item.name}</div>)

// After (fixed):
items.map((item) => <div key={item.id}>{item.name}</div>)
```

For data that genuinely has no unique `id` field, derive a stable key from the item's content (e.g., a composite of its unique fields), or restructure the data source to include IDs.

## Suggested Approach

1. Open each affected file listed above
2. For each `.map()` call using `key={index}`, check if the iterated data has a unique `id` or `_id` field
3. If a stable unique field exists, replace `key={index}` with `key={item.id}` (or whatever the field is)
4. If no unique field exists, consider adding one to the data model or using a combination of fields that together are unique
5. Run the linter to verify no new warnings are introduced

## Acceptance Criteria

- [ ] No `key={index}` patterns remain in mapped lists
- [ ] List re-renders correctly preserve component state
- [ ] Linter passes with zero new warnings

AnushKamble:
## Description

The chat messages in the Sessions page were using the array index as the React key for mapped list items. This can cause issues with React's reconciliation process, leading to unnecessary re-renders and potential state bugs when messages are reordered, filtered, or updated via realtime subscriptions.

## Issue Cause

The `messages.map()` callback was using the second parameter (`index`) as the `key` prop instead of a stable identifier from the data itself.

## What We Did

Changed `key={index}` to `key={msg.id}` in the messages map. The `id` field comes from the Supabase `messages` table primary key, which is guaranteed to be unique and stable across renders. Also removed the unused `index` parameter from the map callback since it is no longer referenced.

## Additional Context

The `messages` state is populated via `.select("*")` on the Supabase `messages` table and through realtime INSERT subscriptions, both of which provide a unique `id` field for every message object.

## Checklist

- [x] Code compiles and runs without errors
- [x] No console warnings about missing keys
- [x] Messages render correctly with stable identity

Fixes #220

### DIFF
--- a/src/pages/Sessions.tsx
+++ b/src/pages/Sessions.tsx
@@ -568,13 +568,13 @@ const Sessions = () => {
 
             {/* MESSAGES */}
             <div className="flex-1 overflow-y-auto py-5 space-y-4">
-              {messages.map((msg, index) => {
+              {messages.map((msg) => {
                 const isCurrentUser =
                   msg.user_id === user?.id;
 
                 return (
                   <div
-                    key={index}
+                    key={msg.id}
                     className={`flex ${
                       isCurrentUser
                         ? "justify-end"


AnushKamble:
Hi @durdana3105, could you please add the relevant labels (like `advanced`, `critical`, `exceptional`) to this PR if it qualifies? This is a best-practices improvement that fixes React rendering performance and prevents potential state bugs with realtime chat messages. Thanks!

PR #222, opened 2026-05-29
fix(resources): Secure uploadResource to prevent IDOR

### LINKED ISSUE #212: Insecure Direct Object Reference (IDOR) in File Uploads
Location: src/lib/uploadResource.ts Description: The uploadResource function accepts a userId argument directly from the client side and uses it to construct the storage file path (${userId}/...) and insert the database record (uploaded_by: userId). Because it completely trusts the client input instead of securely fetching the authenticated user's ID via supabase.auth.getUser() (which is correctly done in deleteResource.ts), a malicious user could tamper with the request and upload files on behalf of another user or an admin. Impact: Broken Access Control / IDOR. Attackers can pollute other users' workspaces, exhaust their storage quotas, or frame them for uploading malicious/illegal content.

riddhima25bet10005-a11y:
Fixes #212

Replaced client-provided userId with secure supabase.auth.getUser() on the backend to fix the IDOR vulnerability.

### DIFF
--- a/src/components/resources/UploadDialog.tsx
+++ b/src/components/resources/UploadDialog.tsx
@@ -90,8 +90,7 @@ const UploadDialog = ({ open, onOpenChange, onSuccess }: UploadDialogProps) => {
       file,
       title.trim(),
       description.trim(),
-      parsedTags,
-      currentUser.id
+      parsedTags
     );
 
     setIsUploading(false);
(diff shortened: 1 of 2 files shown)


riddhima25bet10005-a11y:
Hey @durdana3105 
Everything looks clean.
Kindly review and merge.

PR #229, opened 2026-05-29
fix: restrict global chat components to authenticated users & fix build errors

### LINKED ISSUE #228: [Bug]: Global chat components (Chatbot & FloatingAI) are visible to unauthenticated users
## Overview

Currently, the `<Chatbot />` and `<FloatingAI />` components are rendered globally at the root level inside `App.tsx`.

Because these components do not conditionally verify authentication state and are not mounted inside a protected layout, they remain visible to unauthenticated users across public routes such as:

- Landing page (`/`)
- Login page (`/login`)
- Signup page (`/signup`)

This exposes authenticated-only UI elements to anonymous visitors.

## Steps to Reproduce

1. Open the application in an incognito/private browser window (ensure you are logged out).
2. Navigate to one of the public routes:
   - `/`
   - `/login`
   - `/signup`
3. Observe that the floating AI/chat action buttons are visible.
4. Attempt to open the widget and send a message.

## Actual Behavior

The AI widgets are rendered for unauthenticated users on public pages.

When an anonymous user interacts with them:

- Requests may fail silently
- API calls may throw console errors
- Components attempt authenticated behavior without an active user session

## Expected Behavior

The AI assistant widgets should only be visible to authenticated users with an active session.

Ideally, these components should be mounted exclusively within:

- Dashboard views
- Protected routes
- Authenticated application layouts

They should not render on public pages.

## Proposed Solution

There are multiple viable approaches to resolve this issue.

### Recommended Approach

Move the `<Chatbot />` and `<FloatingAI />` components out of the global root scope in `App.tsx`.

Instead, mount them inside authenticated layouts such as:

- `<WithNav>`
- `<ProtectedRoute>`

This ensures they are only rendered when a valid authenticated session exists.

### Alternative Approach

Update both component implementations to perform authentication checks internally.

For example, use the `useAuth` hook and return early when no authenticated user exists:

```tsx
const { user } = useAuth();

if (!user) {
  return null;
}
```

This prevents rendering for anonymous visitors even if the components remain globally mounted.

## Affected Files

- `src/App.tsx`
- `src/components/Chatbot.jsx`
- `src/components/FloatingAI.tsx`

ArshVermaGit:
## Overview
This PR addresses an issue where global chat widgets (`<Chatbot />` and `<FloatingAI />`) were visible to unauthenticated users on public pages (e.g., the Landing and Login pages). Additionally, it resolves a couple of build-breaking errors encountered during the frontend compilation.

## Resolved Issue
Resolves #228 

## Changes Made
- **Restricted Chat Components**: Modified `src/App.tsx` to conditionally render `<Chatbot />` and `<FloatingAI />` only when an active `user` session exists.
- **Context Fix**: Moved `<FloatingAI />` from the outer `<App />` wrapper into `<AppContent />` so it correctly inherits the authentication context from `AuthProvider`.
- **JSX Syntax Fix (Sessions.tsx)**: Relocated a misplaced `useEffect` hook (used for the User Activity Tracker) out of the JSX return block and into the main component body, resolving a fatal build error.
- **Dependency Fix**: Installed `axios` as a dependency, which was missing from `package.json` but actively used in `src/pages/LearnerDashboard.tsx`.

## Steps to Verify
1. Open the application in an incognito window.
2. Verify that the Floating AI and Chatbot buttons are **not** visible on the Landing page, Login page, or Signup page.
3. Log into an account.
4. Verify that the Chat widgets reappear globally across the application for the authenticated user.
5. Run `npm run build` and ensure the frontend successfully compiles without JSX or missing dependency errors.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,9 +48,11 @@
         "crypto": "^1.0.1",
         "date-fns": "^3.6.0",
         "embla-carousel-react": "^8.6.0",
+        "express": "^4.21.2",
         "framer-motion": "^11.18.2",
         "input-otp": "^1.4.2",
         "lucide-react": "^0.462.0",
+        "mongoose": "^8.18.0",
         "next-themes": "^0.3.0",
         "nodemailer": "^8.0.5",
         "openai": "^6.34.0",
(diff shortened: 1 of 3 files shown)


ArshVermaGit (commented):
Hi @durdana3105 ! Issue #228 has been resolved. Please review the PR and merge it under GSSoC. Thanks!

PR #237, opened 2026-05-29
Fix: Unmodified Profile Skills Wiper (Data Loss)

### LINKED ISSUE #224: Unmodified Profile Skills Wiper (Data Loss)
Location: src/pages/EditProfile.jsx & src/pages/Profile.tsx (Handle Save) Issue: When a user's profile is fetched from the database, profile.skills is an array of strings (e.g., ["React", "Node"]). When the user clicks "Save" without explicitly modifying the skills input field, the save handler evaluates typeof profile.skills === 'string'. Since it's an array, this check fails, causing the fallback logic to set it to an empty array []. This completely wipes out the user's previously saved skills in the database.

riddhima25bet10005-a11y:
Fixes #224. This PR handles the case where profile.skills might already be an array in EditProfile.jsx and Profile.tsx, preventing data loss when saving an unmodified profile.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,9 +48,11 @@
         "crypto": "^1.0.1",
         "date-fns": "^3.6.0",
         "embla-carousel-react": "^8.6.0",
+        "express": "^4.21.2",
         "framer-motion": "^11.18.2",
         "input-otp": "^1.4.2",
         "lucide-react": "^0.462.0",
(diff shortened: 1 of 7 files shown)


riddhima25bet10005-a11y:
Hey @durdana3105 
Everything looks good.
Kindly review and approve.

PR #238, opened 2026-05-29
Fix: Client-Side XP Forgery (Privilege Escalation)

### LINKED ISSUE #225: Client-Side XP Forgery (Privilege Escalation)
Location: src/hooks/useAwardXP.ts Issue: The XP to award is evaluated locally on the client and passed directly into a remote procedure call: await supabase.rpc("increment_user_xp", { _amount: xpToAward }). A malicious user could simply open their browser console and execute this RPC with _amount: 1000000 to instantly level up and unlock all badges. Fix idea: The RPC should instead accept an activity_type string, and the Postgres function should look up the corresponding XP amount securely on the server side.

riddhima25bet10005-a11y:
Fixes #225. This PR introduces a secure server-side RPC (award_activity_xp) that calculates XP based on the activity type rather than trusting the client-provided amount. It also revokes direct execution access to increment_user_xp from authenticated users to prevent arbitrary XP grants.

### DIFF
--- a/src/hooks/useAwardXP.ts
+++ b/src/hooks/useAwardXP.ts
@@ -18,10 +18,8 @@ export const useAwardXP = () => {
 
       const xpToAward = getXPForActivity(activity);
 
-      // Securely delegate to the ledger-backed RPC function.
-      // This completely solves the race condition and updates both the profiles
-      // and leaderboard tables atomically.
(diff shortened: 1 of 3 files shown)


riddhima25bet10005-a11y:
Hey @durdana3105
Everything looks good.
Kindly review and approve.

PR #249, opened 2026-05-29
fix: search input writes to message state instead of search state in Sessions.tsx

### LINKED ISSUE #234: CRITICAL: Search input writes to message state instead of search state in Sessions.tsx -- session search completely dead
## Description

In `src/pages/Sessions.tsx` at line 284, the search input onChange handler writes to the `message` state variable instead of the `search` state variable. This means typing in the search box never updates the search value, and the session filtering useEffect (lines 89-114) always receives an empty string.

## Issue Cause

A copy-paste error: the onChange handler calls `setMessage(e.target.value)` when it should call `setSearch(e.target.value)`. The `message` state (line 45) is intended for the chat message input at line 709, while `search` state (line 53) is intended for the search filter.

## How I Would Fix It

Change `setMessage(e.target.value)` to `setSearch(e.target.value)` on line 284. Also remove the `setIsTyping(true)` and associated setTimeout from the search handler since the typing indicator is meant for chat messages, not search keystrokes.

## Additional Context

- Typing in the search box corrupts the chat message input because it writes to `message` state.
- The typing indicator fires incorrectly for search keystrokes instead of only chat keystrokes.
- Users cannot filter sessions at all -- a core feature of the Sessions page is entirely non-functional.
- Every keystroke in the search box also triggers `setIsTyping(true)`, causing a false "Someone is typing" indicator.

AnushKamble:
## Description

The search input in Sessions.tsx contains a critical copy-paste bug: its onChange handler writes to the `message` state variable (`setMessage(e.target.value)`) instead of the `search` state variable (`setSearch(...)`). Every keystroke in the search box corrupts the chat message input and the session search filtering never works.

Fixes #234:
- Search filtering completely dead (filtering effect always receives empty string)
- Chat message input corrupted by search keystrokes
- Typing indicator fires incorrectly for search keystrokes

## Type of change

- Bug fix (non-breaking change which fixes an issue where search writes to wrong state)
- Code quality (removes dead typing-indicator logic from search handler)

## Root Cause Analysis

`src/pages/Sessions.tsx:348` (before fix):

```tsx
onChange={(e) => {
  setMessage(e.target.value);  // BUG: writes to `message` state instead of `search`
  setIsTyping(true);
  setTimeout(() => {
    setIsTyping(false);
  }, 1500);
}}
```

The `search` state (line 53) is declared alongside `setSearch`, but the onChange calls `setMessage` (line 45). The useEffect at lines 96-114 that filters sessions based on `search` always receives an empty string because `search` is never updated by the input.

## Changes Made

1. **Line 348**: Changed `setMessage(e.target.value)` to `setSearch(e.target.value)`
2. **Lines 350-354**: Removed `setIsTyping(true)` and associated setTimeout -- the typing indicator is meant for the chat message input, not search keystrokes

## Impact

| Scenario | Before | After |
|----------|--------|-------|
| Type in search box | Search filtering dead, chat input corrupted | Search filters correctly, chat input untouched |
| Typing indicator | Fires on every search keystroke | Only fires on chat message changes |

## Checklist

- [x] Search filtering works correctly after fix

### DIFF
--- a/src/pages/Sessions.tsx
+++ b/src/pages/Sessions.tsx
@@ -345,13 +345,7 @@ const [summaryLoading, setSummaryLoading] =
             placeholder="Search sessions..."
             value={search}
             onChange={(e) => {
-            setMessage(e.target.value);
-
-            setIsTyping(true);
-
-            setTimeout(() => {
-              setIsTyping(false);
-            }, 1500);
+            setSearch(e.target.value);
           }}
             className="w-full bg-white/5 border border-white/10 backdrop-blur-xl rounded-2xl py-4 pl-14 pr-4 outline-none focus:border-cyan-400 transition"
           />


AnushKamble:
Hi @durdana3105, could you please add the relevant labels (like `critical`) to this PR if it qualifies? This fixes a bug where the search input writes to the wrong state, making session search completely non-functional. Thanks!

PR #256, opened 2026-05-29
fix: prevent Discover page crash when skills is stored as array instead of string

### LINKED ISSUE #235: CRITICAL: Discover page crashes when searching or filtering if user skills is stored as an array -- TypeError: u.skills.toLowerCase is not a function
## Description

In `src/pages/Discover.tsx`, lines 132-134 (search filter) and lines 141-143 (category filter), the code calls `u.skills?.toLowerCase()` directly on the skills field. If `skills` is stored as a PostgreSQL text array (`TEXT[]`), Supabase returns it as a JavaScript array, and calling `.toLowerCase()` on an array throws a `TypeError`.

## Issue Cause

The `getMatchScore` function (lines 101-103) correctly handles both string and array formats by using `Array.isArray()` checks, but the filter effect bypasses this safe handling and calls `.toLowerCase()` directly on `u.skills` without checking its type.

## How I Would Fix It

Normalize `u.skills` to a string before calling `.toLowerCase()` in both filter blocks, matching the pattern used in `getMatchScore`. For the search filter, join the array with spaces if it is an array. For the category filter, map over the array if it is an array and use `.some()` to check for matches.

## Additional Context

- This is a runtime TypeError that crashes the entire Discover page whenever a user types a search term or selects a filter chip.
- The crash is not caught at compile time because the codebase uses `any` types for the `users` state.
- The crash breaks the "Find Peers" feature entirely -- users see a blank or white screen.
- Because the filter effect has `users` in its dependency array (line 156), this crash occurs on every data fetch.

AnushKamble:
## Description

The Discover page crashes with a `TypeError: u.skills.toLowerCase is not a function` when a user searches or applies a filter and any user in the database has `skills` stored as a PostgreSQL text array (returned as a JavaScript array by Supabase). This breaks the core "Find Peers" feature entirely.

Fixes #235:
- Runtime TypeError crash on every search/filter interaction
- Blank or white screen instead of filtered results
- Entire Discover page unusable for users with array-typed skills

## Type of change

- Bug fix (non-breaking change which fixes a runtime TypeError crash)
- Code quality (aligns filtering logic with the safe pattern already used in `getMatchScore`)

## Root Cause Analysis

`src/pages/Discover.tsx:132-134` and `:141-143` (before fix):

```tsx
// SEARCH filter
matched.filter((u) =>
  u.skills?.toLowerCase().includes(search.toLowerCase())  // CRASH if skills is array
);

// CATEGORY filter
matched.filter((u) =>
  u.skills?.toLowerCase().includes(selectedFilter.toLowerCase())  // CRASH if skills is array
);
```

The `getMatchScore` function (lines 101-103) correctly uses `Array.isArray()` to handle both string and array formats, but the filter effect bypasses this safe handling entirely.

## Changes Made

1. **Search filter (lines 127-136)**: Normalize `u.skills` to a string by joining the array with spaces before calling `.toLowerCase()`, matching the safe pattern from `getMatchScore`.

2. **Category filter (lines 139-149)**: Convert `u.skills` to a string array regardless of stored format, then use `.some()` for partial matching against the selected filter.

## Impact

| Scenario | Before | After |
|----------|--------|-------|
| Search with string-typed skills | Works | Works |
| Search with array-typed skills | CRASH | Works correctly |
| Filter with string-typed skills | Works | Works |
| Filter with array-typed skills | CRASH | Works correctly |

## Checklist

- [x] Discover page works with both string and array skill formats

### DIFF
--- a/src/pages/Discover.tsx
+++ b/src/pages/Discover.tsx
@@ -124,24 +124,27 @@ const Discover = () => {
 
     // SEARCH
     if (search) {
-      matched = matched.filter(
-        (u) =>
-          u.name
-            ?.toLowerCase()
-            .includes(search.toLowerCase()) ||
-          u.skills
-            ?.toLowerCase()
-            .includes(search.toLowerCase())
-      );
+      matched = matched.filter((u) => {
+        const skillsStr = Array.isArray(u.skills)
+          ? u.skills.join(" ").toLowerCase()
+          : (u.skills || "").toLowerCase();
+        return (
+          u.name?.toLowerCase().includes(search.toLowerCase()) ||
+          skillsStr.includes(search.toLowerCase())
+        );
+      });
     }
 
     // FILTERS
     if (selectedFilter !== "All") {
-      matched = matched.filter((u) =>
-        u.skills
-          ?.toLowerCase()
-          .includes(selectedFilter.toLowerCase())
-      );
+      matched = matched.filter((u) => {
+        const skillsList = Array.isArray(u.skills)
+          ? u.skills.map((s: string) => s.toLowerCase())
+          : (u.skills?.split(",") || []).map((s: string) => s.trim().toLowerCase());
+        return skillsList.some((skill: string) =>
+          skill.includes(selectedFilter.toLowerCase())
+        );
+      });
     }
 
     // DEFAULT BUBBLE


AnushKamble:
Hi @durdana3105, could you please add the relevant labels (like `critical`) to this PR if it qualifies? This fixes a runtime TypeError crash on the Discover page when user skills are stored as an array instead of a string. Thanks!

PR #313, opened 2026-05-29
Fixes #290: Implement DB insert rate limiting for chat

### LINKED ISSUE #290: Absence of Database Insertion Rate Limiting for Chat
Description: While Auth and AI endpoints have rate limits, standard API inserts into messages and study_room_messages rely only on RLS. Impact: A malicious authenticated user can write a loop to insert 10,000+ messages per minute. Supabase does not rate limit API inserts natively, meaning the attacker can easily bloat the database storage, consume all API egress quotas, and increase cloud billing costs drastically.

riddhima25bet10005-a11y:
## Description
This PR resolves the missing database insertion rate limit vulnerability identified in Issue #290. 

### The Problem
Previously, inserting messages via the Supabase PostgREST API relied solely on Row Level Security (RLS). While RLS ensures users can only insert messages on their own behalf, there was no limit on the *rate* of insertion. A malicious authenticated user could easily craft a simple script to spam 10,000+ inserts per minute. Because Supabase API does not apply rate limiting natively to standard inserts, this would severely bloat database storage, exhaust API egress quotas, and potentially increase cloud computing costs dramatically.

### The Solution
To fix this vulnerability, this PR introduces hard rate limiting at the Postgres database level:
1. **Rate Limit Function:** Created a PL/pgSQL function (`check_message_rate_limit`) that intercepts `INSERT` commands. It dynamically calculates the number of messages a user has sent in the last 10 seconds.
2. **Threshold Enforcement:** If a user attempts to insert 5 or more messages within a 10-second rolling window, the function throws a Postgres exception (`RAISE EXCEPTION 'Rate limit exceeded...'`). This immediately blocks the insertion and returns an error response to the client via PostgREST.
3. **Database Triggers:** Applied `BEFORE INSERT` triggers to both the `messages` table (used for Direct Messages) and the `study_room_messages` table (used for group chats). The trigger automatically checks the correct user ID column (`sender_id` vs `profile_id`) depending on the context.

This solution entirely blocks malicious insertion loops without requiring any modifications to frontend logic.

Fixes #290

### DIFF
--- a/supabase/migrations/20260529000005_chat_rate_limiting.sql
+++ b/supabase/migrations/20260529000005_chat_rate_limiting.sql
@@ -1,0 +1,54 @@
+-- Create a robust function to check rate limits for message inserts
+CREATE OR REPLACE FUNCTION check_message_rate_limit()
+RETURNS trigger AS $$
+DECLARE
+  v_message_count INT;
+  v_user_id UUID;
+BEGIN
+  -- Identify the user based on the table structure
+  IF TG_TABLE_NAME = 'messages' THEN
+    v_user_id := NEW.sender_id;
+  ELSIF TG_TABLE_NAME = 'study_room_messages' THEN
+    v_user_id := NEW.profile_id;
+  END IF;
+
+  -- Ensure we have a user_id to check against
+  IF v_user_id IS NULL THEN
+    RETURN NEW;
+  END IF;
+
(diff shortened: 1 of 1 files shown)


riddhima25bet10005-a11y:
Hey @durdana3105
Kindly review and approve.

PR #315, opened 2026-05-29
Fixes #292 - Enforce Future scheduled_at for Sessions

### LINKED ISSUE #292: Logic Flaw: Time-Travel Session Creation
Description: The frontend validation requires scheduled_at to be at least 1 hour in the future. However, there is no corresponding database CHECK constraint ensuring scheduled_at >= now(). Impact: An attacker can create a session via the API scheduled for the year 1990. The tick_session_statuses() function will immediately process it, marking it as ended. This can be abused to artificially inflate a mentor's "completed sessions" metrics and abuse the gamification/XP system without actually hosting a session.

riddhima25bet10005-a11y:
## Description
Fixes #292 by enforcing a strict validation on `scheduled_at` at the database level. Previously, an attacker could bypass frontend validation and create a session scheduled in the past (e.g., year 1990) via the API, which the system would instantly flag as "completed", allowing malicious users to quickly farm XP and artificially inflate their "completed sessions" metrics.

## Changes Made
- Added a new database migration (`20260529000006_enforce_future_scheduled_at.sql`) that introduces a PL/pgSQL function `check_future_scheduled_at()`.
- Added a `BEFORE INSERT OR UPDATE` trigger on the `sessions` table using the above function.
- The trigger strictly ensures that `scheduled_at` must be in the future (with a 5-minute clock-drift grace period) anytime a session is inserted or its schedule is modified.

## Gamification Protection
This permanently patches the vulnerability where users could exploit the `tick_session_statuses()` chron logic to gain unauthorized XP by time-traveling their sessions into the past.

## GSSoC '26 Guidelines
- PR Title should be: `Fixes #292 - Enforce Future scheduled_at for Sessions`
- The issue is linked so it automatically closes upon merge.

### DIFF
--- a/supabase/migrations/20260529000006_enforce_future_scheduled_at.sql
+++ b/supabase/migrations/20260529000006_enforce_future_scheduled_at.sql
@@ -1,0 +1,26 @@
+-- Create function to enforce that a session's scheduled_at is in the future
+CREATE OR REPLACE FUNCTION check_future_scheduled_at()
+RETURNS trigger AS $$
+BEGIN
+  -- Only validate if scheduled_at is being inserted or modified
+  IF NEW.scheduled_at IS NOT NULL THEN
+    -- If it's an INSERT, or an UPDATE where scheduled_at actually changed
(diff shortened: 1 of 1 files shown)


riddhima25bet10005-a11y:
Hey @durdana3105
Kindly review and approve.

PR #326, opened 2026-05-30
Added auth/authorization checks + rate limiting + secure email flows

### LINKED ISSUE #299: Strengthen security: auth/authorization checks + rate limiting + secure email flows
Description

Ensure every protected backend route verifies authorization (not only authentication).
Mitigate abuse on endpoints like login/reset-password/OTP/email sending.
Harden email verification/reset flows against enumeration.
Implementation plan

Backend (backend/routers, backend/controllers):
Review each route to ensure user identity is enforced (e.g., mentor/learner roles) before returning data.
Add rate limiting middleware (e.g., express-rate-limit) for:
login
signup
forgot/reset password
OTP verification
Prevent user enumeration:
Always return the same success message for “email sent” even if user doesn’t exist.
Ensure secure tokens:
short expiry
single-use where possible
store hashes, not raw tokens

Aditya8369:
Implemented the backend hardening in both trees the repo currently contains. The active server.js path now has Mongo startup, reusable rate limiting, and Supabase profile-based authorization on protected routes via requireAuth.js and rateLimiter.js. The password reset flow in authController.js now uses hashed single-use tokens, a 15-minute expiry, generic success messaging, and token cleanup on mail failure. I also applied the same hardening to the duplicate top-level backend files under server.js, authController.js, and related routers so the weaker copy is not left behind.

Validation passed with syntax checks on both backend trees. 

closes #299

### DIFF
--- a/backend/controllers/authController.js
+++ b/backend/controllers/authController.js
@@ -1,27 +1,133 @@
-// forgotPassword
-// resetPassword
+import bcrypt from "bcryptjs";
+import crypto from "crypto";
+import User from "../models/User.js";
+import { sendEmail } from "../utils/sendEmail.js";
 
-// backend/controllers/authController.js
+const RESET_TOKEN_TTL_MS = 15 * 60 * 1000;
+const GENERIC_RESET_MESSAGE =
+  "If an account with that email exists, a password reset link has been sent.";
+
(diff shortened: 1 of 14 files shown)


Aditya8369:
@durdana3105 please check this one

PR #357, opened 2026-05-30
fix(frontend): replace direct DOM manipulation with React state for sparkle effect

### LINKED ISSUE #349: [Bug] Direct DOM Manipulation in React App (Memory Leak Risk)
## 📝 Description
In `src/App.tsx`, there is a `useEffect` hook that listens to the `mousemove` event and manually creates DOM elements (`document.createElement('div')`) to create a "sparkle" trail effect. It then injects these elements directly into the document using `document.body.appendChild`.
This is a severe anti-pattern in React. React manages the DOM via the Virtual DOM (VDOM). When we bypass the VDOM to manually create and append elements, React loses track of them. Furthermore, since the `mousemove` event fires dozens of times per second, this causes heavy layout thrashing, DOM bloating, and severe memory leaks, which will progressively slow down the application the longer a user moves their mouse.
## 💥 Impact
- **Performance Degradation:** Continuous direct DOM appending on rapid events like `mousemove` causes forced synchronous layouts and frame drops.
- **Memory Leaks:** If the manually created elements are not properly cleaned up or removed, they will pile up endlessly in the DOM, eventually crashing the user's browser tab.
- **React Reconciliation Conflicts:** Direct DOM manipulation can interfere with React's rendering lifecycle and potentially crash the application if React attempts to modify a node it didn't create.
## ✅ Proposed Solution (The Fix)
1. **Remove Direct DOM API Usage:** Delete the `document.createElement` and `appendChild` logic from `src/App.tsx`.
2. **Implement React State:** Refactor the effect to use React state. Maintain an array of objects containing `x`, `y`, and `id` properties for the sparkles. Use the `mousemove` handler to push new coordinate objects into this state (with a debouncer or throttle to save performance) and automatically prune old ones using `setTimeout`.
3. **Map State to Components:** Render the sparkles safely inside the JSX return block by mapping over the state array (e.g., `{sparkles.map(s => <div key={s.id} className="sparkle" ... />)}`).
4. **Alternative:** Abstract this visual flair entirely by using a dedicated animation library like `framer-motion` or `tsparticles`, which are heavily optimized for React environments.
## 📂 Files Involved
- `src/App.tsx`
- `src/index.css` (If styles need adjustment)
## 🛠️ Steps to Reproduce
1. Start the frontend application.
2. Open your browser's Developer Tools (Elements tab).
3. Move your mouse rapidly around the screen.
4. Observe the Elements tab—you will see hundreds of `div.sparkle` elements being appended to the `<body>` outside of the React `<div id="root">` container.

ArshVermaGit:
## 📝 Context & Motivation
The application featured a "sparkle" trail effect on mouse movement in `src/App.tsx`. However, the implementation was using native DOM APIs (`document.createElement` and `appendChild`) to inject raw `div` elements into the `<body>` outside of the React lifecycle. Because the `mousemove` event fires rapidly, this caused severe layout thrashing, DOM bloating, and memory leaks that degraded application performance over time.

## 🛠️ Changes Implemented
- **Removed Native DOM APIs:** Completely removed all usage of `document.createElement`, `.appendChild`, and `.remove()` from the `mousemove` event listener.
- **Implemented React State:** Refactored the `sparkle` trail to be fully managed by the React Virtual DOM using a `useState` array to store active sparkle coordinates.
- **Safe Cleanup:** Implemented self-cleaning timeouts that safely filter sparkles out of the React state array after 800ms, effectively resolving the memory leak.
- **Component Rendering:** React now declaratively maps over the state array to render the `.sparkle` elements natively within the JSX tree.

## 🧪 How to Test
1. Start the frontend development server (`npm run dev`).
2. Open the browser and move your mouse around the screen to trigger the sparkle effect.
3. Open Developer Tools (Elements tab) and verify that the `div.sparkle` elements are now safely rendering *inside* the React `#sparkle-container` rather than being appended directly to the body.
4. Keep moving the mouse and watch the DOM tree to verify that old sparkles are successfully unmounting after 800ms, preventing infinite DOM bloat.

## 🔗 Related Issues
- Resolves #349

### DIFF
--- a/backend/controllers/aiController.js
+++ b/backend/controllers/aiController.js
@@ -1,121 +1,92 @@
-export const askAI = async (req, res) => {
-  try {
-    const { question } = req.body;
+import { asyncHandler } from "../utils/asyncHandler.js";
 
-    if (!question || typeof question !== "string") {
-      return res.status(400).json({ error: "Invalid question provided" });
+export const askAI = asyncHandler(async (req, res) => {
+  const { question } = req.body;
+
+  const response = await fetch(
+    "https://openrouter.ai/api/v1/chat/completions",
+    {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        Authorization: `Bearer ${process.env.OPENROUTER_API_KEY}`,
+      },
+      body: JSON.stringify({
+        model: "openai/gpt-4o-mini",
+        messages: [
+          {
+            role: "system",
+            content:
+              "You are an AI peer mentor for students. Answer questions about coding, AI, DSA, and roadmaps in a supportive, clear, and approachable way.",
+          },
+          {
+            role: "user",
+            content: question,
+          },
+        ],
+      }),
     }
+  );
 
-    if (question.length > 2000) {
-      return res.status(400).json({ error: "Question exceeds maximum length of 2000 characters" });
-    }
+  const data = await response.json();
 
-    const response = await fetch(
-      "https://openrouter.ai/api/v1/chat/completions",
-      {
-        method: "POST",
-        headers: {
-          "Content-Type": "application/json",
-          Authorization: `Bearer ${process.env.OPENROUTER_API_KEY}`,
-        },
-        body: JSON.stringify({
-          model: "openai/gpt-4o-mini",
-          messages: [
-            {
-              role: "system",
-              content:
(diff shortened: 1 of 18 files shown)


ArshVermaGit (commented):
Hi @durdana3105 ! The issue assigned to me has been resolved in PR #349  . Please consider reviewing it and merging it under GSSoC. Thanks!

PR #383, opened 2026-05-31
[ENHANCE] Improve AI route robustness: rate limiting, token budgeting, and streaming-safe responses

### LINKED ISSUE #337: [ENHANCE] Improve AI route robustness: rate limiting, token budgeting, and streaming-safe responses
Description
AI endpoints often need stricter controls: rate limits per user/IP, max prompt size, timeouts, and safe handling when OpenRouter/LLM fails. This reduces cost spikes and user-facing failures.

Implementation steps

Inspect backend/controllers/aiController.js and backend/routers/aiRoutes.js.
Enhance rate limiter middleware to use a stable key (user id if available, else IP) and return a consistent 429 payload.
Add request constraints:
max message length
max number of messages in context
max overall prompt size
Add timeout handling around the OpenRouter call (e.g., abort fetch after X ms).
Add “fallback responses” (return a helpful error + retry suggestion) when the model call fails.
Ensure logs don’t leak secrets; log only request ids / user ids and high-level failure reasons

Aditya8369:
AI endpoints are hardened in aiController.js, openRouter.js, aiRoutes.js, and chatRoutes.js. Requests now get tighter schema bounds on message count, message length, and total prompt size; AI traffic uses a dedicated per-user-or-IP limiter; OpenRouter calls are time-bounded; and failures return a consistent 503 with retry guidance instead of raw upstream errors.

Logging is now request-scoped and high-level only via app.js and errorHandler.js, and I removed the startup env-value logs from server.js. I also added coverage in aiRobustness.test.js alongside the existing validation tests in validation.test.js.

Verified with Vitest and ESLint; both passed. If the src/backend mirror is still active in your app, I can sync the same changes there next.

closes #337

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -1,4 +1,5 @@
 import express from "express";
+import { randomUUID } from "crypto";
 import authRoutes from "./routers/authRoutes.js";
 import chatRoutes from "./routers/chatRoutes.js";
 import aiRoutes from "./routers/aiRoutes.js";
@@ -8,6 +9,11 @@ const app = express();
 
 app.set("trust proxy", 1);
 app.use(express.json());
+app.use((req, res, next) => {
+	req.requestId = req.headers["x-request-id"] || randomUUID();
+	res.setHeader("x-request-id", req.requestId);
+	next();
+});
 
 app.use("/api/ai", aiRoutes);
 app.use("/api", authRoutes);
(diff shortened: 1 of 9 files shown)


durdana3105:
this pr have merge conflicts please resolve it

Aditya8369:
@durdana3105 solved conflicts

PR #394, opened 2026-05-31
Fixes #373: Arbitrary Data Insertion in skills_taxonomy

### LINKED ISSUE #373: Arbitrary Data Insertion in skills_taxonomy
Location: supabase/migrations/20260530000001_skills_taxonomy.sql The RLS policy Authenticated users can insert skills is defined with WITH CHECK (true) and lacks backend format validation. Any authenticated user can insert thousands of unvalidated rows (or even garbage/offensive text) into the global skills_taxonomy table, disrupting the application's filtering and taxonomy systems.

riddhima25bet10005-a11y:
This PR fixes issue #373 by adding backend format validation to the \skills_taxonomy\ table. A \CHECK\ constraint has been introduced to limit the maximum length of a skill name to 50 characters, and requires that the skill name contains only valid characters (alphanumeric, spaces, and specific symbols like ., -, +, #, /). Any existing data that violates these rules is removed before the constraint is applied. This prevents arbitrary, unbounded, or offensive data from being inserted by authenticated users.

### DIFF
--- a/supabase/migrations/20260531000003_fix_skills_taxonomy.sql
+++ b/supabase/migrations/20260531000003_fix_skills_taxonomy.sql
@@ -1,0 +1,16 @@
+-- Fix Arbitrary Data Insertion in skills_taxonomy by adding backend format validation
+
+-- Remove any existing invalid data to allow the constraint to be applied
+DELETE FROM public.skills_taxonomy
+WHERE length(name) > 50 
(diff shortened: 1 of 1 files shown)


riddhima25bet10005-a11y:
Hey @durdana3105 
Everything looks good.
Kindly review and merge.

PR #395, opened 2026-05-31
Fixes #374: Broken Access Control in sessions Creation (Role Bypass)

### LINKED ISSUE #374: Broken Access Control in sessions Creation (Role Bypass)
Location: supabase/migrations/20260518_app_bootstrap_and_notifications.sql The Authenticated users can create sessions policy allows anyone to insert a session as long as mentor_id = auth.uid(). However, the policy fails to verify if the user actually holds the is_mentor = true flag (stored in the profiles table). Consequently, standard learners can bypass the mentor approval workflow to create sessions and pollute the upcoming sessions feed.

riddhima25bet10005-a11y:
This PR fixes issue #374 by hardening the RLS policies for the \sessions\ table. It drops all old permissive or conflicting policies (such as the one from the bootstrap migration that allowed any authenticated user to create sessions) and enforces strict checks on both INSERT and UPDATE operations. Users can now only insert or update sessions if \mentor_id = auth.uid()\ AND they have the \is_mentor = true\ flag in their profile, preventing standard learners from bypassing the workflow.


riddhima25bet10005-a11y:
Hey @durdana3105 
Everything is good.
Kindly review and merge.

PR #433, opened 2026-06-01
fix: Remove redundant auth calls and duplicate Supabase filters in Chatbot.jsx

### LINKED ISSUE #420: fix: Remove redundant auth calls and duplicate Supabase filters in Chatbot.jsx
# Overview

This cleanup addresses leftover merge artifacts in `src/components/Chatbot.jsx` that introduce redundant authentication logic, duplicate query filters, and dead code.

While the component remains functionally operational, the current implementation performs unnecessary Supabase auth requests and contains stale implementation remnants from an unresolved merge.

---

### 1. Redundant Duplicate `.eq()` Filter in `loadChats`

The `loadChats()` query currently applies two identical filters against the same column:

```jsx
const { data } = await supabase
  .from("chat_messages")
  .select("*")
  .eq("user_id", user.id)
  .eq("user_id", session.user.id)
  .order("created_at", { ascending: true });
```

Both values resolve to the same authenticated user ID.

#### Problem

The second filter is completely redundant because:

```txt
user.id === session.user.id
```

This adds unnecessary noise to the query chain and reflects unresolved merge duplication.

#### Expected Fix

Retain only a single user filter:

```jsx
.eq("user_id", session.user.id)
```

or whichever auth source is standardized after cleanup.

---

### 2. Redundant Dual Authentication Calls

`Chatbot.jsx` currently performs both:

```jsx
supabase.auth.getUser()
```

and

```jsx
supabase.auth.getSession()
```

back-to-back.

Example:

```jsx
const { data: { user } } = await supabase.auth.getUser();
if (!user) return;

const { data: { session } } = await supabase.auth.getSession();
if (!session?.user) return;
```

#### Problem

This creates unnecessary duplicate auth requests.

`getSession()` already provides:

- authenticated user object
- access token
- session metadata

Calling `getUser()` immediately beforehand provides no additional benefit.

#### Expected Fix

Replace both calls with a single session retrieval:

```jsx
const {
  data: { session },
} = await supabase.auth.getSession();

if (!session?.user) return;
```

Use:

```jsx
session.user.id
```

throughout the component as the canonical auth source.

---

### 3. Dead Commented-Out Fetch Block

The `sendMessage()` implementation still contains a commented-out legacy `fetch()` request to:

```txt
/api/chat
```

This block remains from an earlier implementation strategy.

#### Problem

The code is:

- unreachable
- commented out
- no longer used by the current flow

Keeping legacy blocks increases maintenance overhead and reduces readability.

#### Expected Fix

Remove the obsolete commented implementation entirely.

---

### 4. Duplicate Auth Pattern Also Exists in `sendMessage()`

The same redundant auth pattern appears in `sendMessage()` (approximately lines 54–60).

Current behavior mirrors `loadChats()` by calling both:

- `getUser()`
- `getSession()`

#### Problem

This duplicates the same unnecessary network overhead during every message submission.

#### Expected Fix

Apply the same cleanup strategy used for `loadChats()`:

- remove `getUser()`
- standardize on `getSession()`
- reference `session.user`

consistently.

---

## Root Cause

These issues appear to originate from a faulty merge resolution where two authentication approaches were concatenated instead of reconciled.

As a result, the component contains overlapping logic from both branches.

---

#### Before

```jsx
.eq("user_id", user.id)
.eq("user_id", session.user.id)
```

#### After

```jsx
.eq("user_id", session.user.id)
```

---

#### Before

```jsx
const { data: { user } } =
  await supabase.auth.getUser();

const { data: { session } } =
  await supabase.auth.getSession();
```

#### After

```jsx
const {
  data: { session },
} = await supabase.auth.getSession();

if (!session?.user) return;
```

---

### Remove Dead Code

Delete the legacy commented `fetch("/api/chat")` implementation from `sendMessage()`.

---

### Standardize Across Component

Apply the same auth cleanup to:

- `loadChats()`
- `sendMessage()`

to maintain consistent patterns across the component.

---

## Impact

This cleanup will:

- reduce unnecessary Supabase auth requests
- simplify query logic
- improve readability
- eliminate merge artifacts
- reduce future maintenance burden
- standardize authentication handling across the component

---

## Files Involved

- `src/components/Chatbot.jsx`

ArshVermaGit:
## Summary

Cleans up merge artifacts in `Chatbot.jsx` that caused redundant network requests and dead code.

## Resolved Issue
Resolves #420 

## Problem

The `Chatbot.jsx` component had leftover code from a previous merge where two branches' auth approaches were concatenated instead of resolved:

1. **Dual auth calls** — Both `supabase.auth.getUser()` and `supabase.auth.getSession()` were called back-to-back in `loadChats` and `sendMessage`. Only one is needed since `getSession()` provides both `session.user` (for the user ID) and `session.access_token` (for the Authorization header).

2. **Duplicate `.eq()` filter** — The `loadChats` query chained `.eq("user_id", user.id)` and `.eq("user_id", session.user.id)` on the same column. Both resolve to the same value, making one completely redundant.

3. **Commented-out dead code** — A 14-line commented-out alternative `fetch()` call to `/api/chat` was left in the `sendMessage` function.

4. **Broken indentation** — The `fetch()` call body was flush-left instead of properly indented inside the `try` block (artifact of the merge).

## Changes Made

- Replaced dual `getUser()` + `getSession()` calls with a single `getSession()` call in both `loadChats` and `sendMessage`
- Removed the duplicate `.eq("user_id", ...)` filter in the chat messages query
- Deleted the commented-out alternative fetch block (lines 92–105)
- Fixed indentation of the `fetch()` call to match the surrounding code

## Impact

- **−2 unnecessary Supabase network requests** per chat load and per message send
- **−28 lines** of dead/redundant code (230 → 202 lines)
- No functional behavior change — the component works exactly the same way

### DIFF
--- a/src/components/Chatbot.jsx
+++ b/src/components/Chatbot.jsx
@@ -23,22 +23,13 @@ export default function Chatbot() {
   // Load only the current user's chat messages
   useEffect(() => {
     const loadChats = async () => {
-
-      const { data: { user } } = await supabase.auth.getUser();
-      if (!user) return;
-
       const { data: { session } } = await supabase.auth.getSession();
       if (!session?.user) return;
-
 
       const { data } = await supabase
         .from("chat_messages")
         .select("*")
-
-        .eq("user_id", user.id)
-
         .eq("user_id", session.user.id)
-
         .order("created_at", { ascending: true });
 
       if (data) setMessages(data);
@@ -49,10 +40,6 @@ export default function Chatbot() {
   // SEND MESSAGE
   const sendMessage = async () => {
     if (!input.trim()) return;
-
-
-    const { data: { user } } = await supabase.auth.getUser();
-    if (!user) return;
 
     const { data: { session } } = await supabase.auth.getSession();
     if (!session?.user) return;
@@ -77,32 +64,17 @@ export default function Chatbot() {
         content: msg.text,
       }));
       const res = await fetch(`${API_BASE_URL}/api/ai/ask`, {
-  method: "POST",
-  headers: {
-    "Content-Type": "application/json",
-    Authorization: `Bearer ${session.access_token}`, // still pass session for user auth
-  },
-  body: JSON.stringify({
-    messages: formattedMessages,
-    systemPrompt: systemPrompt.content,
-    model: "openai/gpt-4", 
-  }),
-});
-
-      // Route the request through the backend so the API key stays server-side.
-      // Include the session token so the backend can authenticate the request.
-      // const res = await fetch("/api/chat", {
-      //   method: "POST",
-      //   headers: {
-      //     "Content-Type": "application/json",
-      //     Authorization: `Bearer ${session.access_token}`,
-      //   },
-      //   body: JSON.stringify({
-      //     messages: formattedMessages,
-      //     systemPrompt: systemPrompt.content,
-      //     model: "openai/gpt-3.5-turbo",
-      //   }),
-      // });
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+          Authorization: `Bearer ${session.access_token}`,
+        },
+        body: JSON.stringify({
+          messages: formattedMessages,
+          systemPrompt: systemPrompt.content,
+          model: "openai/gpt-4",
+        }),
+      });
 
       const data = await res.json();
 


ArshVermaGit (commented):
Hi @durdana3105 ! The issue assigned to me has been resolved in PR #420 . Please consider reviewing it and merging it under GSSoC. Thanks!

PR #470, opened 2026-06-01
fix: remove dead fetch code from FloatingAI, use API_BASE_URL, improve error handling

### LINKED ISSUE #399: AI chat floating widget has broken API calls and incorrect authentication
## Description

The FloatingAI component (`src/components/FloatingAI.tsx`) has broken fetch logic that prevents the AI assistant from working in production. The component contains syntax errors, uses the wrong authentication token format, and duplicates functionality with two other AI chat implementations.

### 1. Broken fetch syntax prevents any API call from completing

The first `fetch()` call has an unclosed object literal (missing `})`), causing a syntax error. The `const response` declaration is then re-declared, which would throw a `SyntaxError: Identifier 'response' has already been declared` in strict mode.

### 2. Wrong authentication token format

```tsx
const token = localStorage.getItem("token");
```

Supabase stores auth tokens under the key `sb-{project-ref}-auth-token`, not `"token"`. The backend's `requireAuth` middleware expects a valid Supabase JWT obtained via `supabase.auth.getSession()`. Using `localStorage.getItem("token")` will always return `null`, causing the backend to return 401 on every request.

### 3. Hardcoded relative URL fails in production

The second fetch uses `/api/chat` (relative URL). In production (deployed via Vercel), there is no Vite dev server proxy, so this request goes to the Vercel deployment's own origin — there's no Express backend there. The request will always 404.

### 4. Three overlapping AI chat implementations

The codebase has three separate AI chat UIs that implement similar functionality with different APIs:
- `src/components/FloatingAI.tsx` — floating widget (broken)
- `src/components/Chatbot.jsx` — embedded chatbot (different endpoint, different styling)
- `src/pages/aipage.tsx` — full-page AI matchmaker (uses Supabase Edge Functions)

This triples maintenance burden and creates inconsistent user experiences.

## Impact

- The FloatingAI widget is completely non-functional in production
- Users get 401 errors (wrong token) or 404 errors (wrong URL) with no helpful feedback
- The widget silently fails — errors are caught and swallowed
- Developer time is wasted maintaining three separate AI chat components

## Files Affected

- `src/components/FloatingAI.tsx` — broken fetch, wrong token, missing error handling
- `src/components/Chatbot.jsx` — duplicate AI chat UI
- `src/pages/aipage.tsx` — third AI chat UI (may be intentionally different)
- `src/contexts/AuthContext.tsx` — source of truth for Supabase session token

## Suggested Fix Direction

1. Fix the syntax error in the fetch call (close the options object and remove duplicate declaration)
2. Replace `localStorage.getItem("token")` with proper Supabase session token from `useAuth()` context
3. Use an environment-aware URL resolver that works in both dev (Vite proxy) and production (absolute backend URL)
4. Consolidate the three AI chat components into a single reusable `AIChatWidget` component with configurable props
5. Add user-facing error toasts when API calls fail instead of silently swallowing errors

AnushKamble:
## Problem

The FloatingAI chat component had a syntax error that prevented compilation and several runtime issues:

1. **Dead/broken code blocks compilation**: Lines 41-46 contained an incomplete const response = await fetch(...) that was never closed, followed by a second const response = await fetch(...) in the same scope - a duplicate declaration syntax error.

2. **Hardcoded relative URL**: The working fetch used '/api/chat' (relative path) instead of the project-wide API_BASE_URL constant.

3. **Poor error handling for auth failures**: A 401 from the backend would hit the generic catch block and show 'AI failed' rather than directing the user to log in.

4. **No differentiation between network/server errors**: The catch block showed the same generic message for all error types.

## What Changed

- **Removed dead code**: Deleted the incomplete, duplicate fetch block that blocked compilation.
- **API_BASE_URL**: Changed the URL from '/api/chat' to ${API_BASE_URL}/api/chat, consistent with SuggestedPartners.tsx.
- **Auth token extraction**: Moved before the fetch call (was buried between dead code and working fetch).
- **Explicit 401 handling**: Shows 'Please log in to use the AI assistant.' instead of cryptic error.
- **Explicit non-ok status handling**: Returns user-friendly message with HTTP status code.
- **Improved catch block**: Shows 'Network error. Please check your connection.' for network failures.
- **Console.error**: Changed from console.log(error) to console.error(...).

## Impact
- The component now compiles without syntax errors.
- Users see a clear 'Please log in' message when their session expires.
- Network errors are distinguished from server errors.
- The component works consistently across dev and production environments.

Closes #399


AnushKamble:
Hi @durdana3105! This PR addresses issue #399 by fixing a compilation-blocking syntax error and several runtime issues in the FloatingAI chat component.

Key changes:
- Removes dead/duplicate fetch code that blocked compilation
- Uses API_BASE_URL instead of hardcoded relative URL
- Adds explicit 401 handling and proper network error messages
- Improves console output (console.log -> console.error)

Please review when you get a chance. The dead code was blocking builds entirely.

Tags: high-priority bug frontend floating-ai compilation

PR #500, opened 2026-06-01
Perf: Isolate Sparkles mousemove effect to fix global React re-rendering

### LINKED ISSUE #480: [Performance]: Severe UI Bottleneck due to Global MouseEvent Sparkles
## 🚨 Description
In `src/App.tsx`, there is a global `mousemove` event listener attached to the window to generate an interactive "sparkles" visual effect tracing the cursor. 
The current implementation leverages React State (`setSparkles`) to track the exact `x` and `y` coordinates of these elements, and pushes new elements into an array on *every single pixel movement* of the user's mouse. 
## 💥 Impact
Because this state array lives inside `AppContent`—which is the absolute root component wrapping the `<Suspense>` router and all page components—every pixel movement forces React to trigger a massive re-render lifecycle for the *entire application DOM tree*. 
- **Latency & Stuttering:** As the DOM gets more complex on pages with charts or large lists, the UI will start stuttering heavily as React struggles to diff the entire app 60+ times per second.
- **Battery Drain:** On mobile devices or laptops, this constant JavaScript execution and DOM repainting will max out the CPU and drain the battery aggressively.
## 👣 Steps to Reproduce
1. Start the frontend application locally (`npm run dev`).
2. Open the Chrome DevTools and navigate to the **Performance** tab (or install the React Developer Tools extension and turn on "Highlight updates when components render").
3. Move your mouse rapidly across the screen for 5 seconds.
4. **Observe:** The React DevTools will show the entire application flashing constantly, and the Performance tab will show massive CPU spikes and dropped frames.
## 💡 Proposed Solutions
Visual effects like this should never trigger global React state updates. We have two options to optimize this:
### Option 1: Isolate into a Native Canvas / WebGL Component (Recommended)
- Remove the `sparkles` state entirely from React.
- Create an isolated `<SparklesCanvas />` component that grabs a raw HTML5 `<canvas>` ref.
- Attach the `mousemove` event listener directly inside this component, drawing the sparkles on the Canvas API natively using `requestAnimationFrame`. This completely bypasses the React rendering engine and runs blazingly fast.
### Option 2: Isolated DOM Manipulation (No React State)
- If we want to stick to DOM elements (divs) instead of a Canvas, create a standalone `<SparkleEffect />` component that does not take props or use `useState`.
- Let it attach the `mousemove` listener, and internally use standard Javascript `document.createElement('div')` to append/remove sparkle divs manually from a wrapper node.
## ✅ Acceptance Criteria
- [ ] The global `setSparkles` state is entirely removed from `App.tsx` and `AppContent`.
- [ ] The entire application no longer re-renders when the mouse moves.
- [ ] The "sparkles" effect still functions properly for the user visually.
- [ ] Mouse movements no longer cause dropped frames in Chrome DevTools Performance profiling.
## 🏷️ Labels
`performance`, `frontend`, `ui`, `tech-debt`, `critical`

ArshVermaGit:
## 🚀 Description
This Pull Request perfectly resolves a severe frontend performance bottleneck that was causing massive CPU spikes, UI stuttering, and dropped frames during user interactions.

Previously, the "sparkles" visual effect tracked exact `x` and `y` mouse coordinates globally using React state (`setSparkles`) inside the root `AppContent` component. Because `AppContent` wraps the entire application router, every single pixel of mouse movement triggered a cascading React re-render of the entire DOM tree!

### 🛠️ Key Changes:
- **State Isolation**: Extracted the `sparkles` state, `useRef`, and `useEffect` event listener out of `App.tsx`.
- **New Component**: Created a standalone `src/components/Sparkles.tsx` component that is wrapped in `React.memo()`. 
- **Impact**: Now, when the user moves their mouse, *only* the isolated `<Sparkles />` component re-renders to draw the elements. The main application router, pages, and heavy UI components remain completely untouched, resulting in a perfectly smooth 60fps experience!

## 🔗 Related Issues
Resolves #480

## 🧪 Testing Instructions
1. Pull this branch (`main_4`) and run `npm run dev`.
2. Open Chrome DevTools and navigate to the **Performance** tab (or turn on "Highlight updates when components render" in the React DevTools).
3. Move your mouse rapidly around the screen.
4. **Verify**: The sparkles follow the cursor perfectly as before, but the rest of the application (navigation, background, content) no longer flashes green in React DevTools, and CPU usage remains flat!

## ✅ Checklist
- [x] Extracted `sparkles` state out of the root `AppContent`.
- [x] Created an isolated `Sparkles.tsx` component.
- [x] Wrapped the new component in `React.memo` to prevent unnecessary upstream renders.


ArshVermaGit (commented):
Hi @durdana3105 ! The issue assigned to me has been resolved in PR #480 . Please consider reviewing it and merging it under GSSoC. Thanks!

PR #562, opened 2026-06-01
Fix race condition (TOCTOU) in join_session RPC (#452)

### LINKED ISSUE #452: Race Condition (TOCTOU) in Session Joining (join_session RPC)
Description: While the join_session RPC uses FOR UPDATE to lock the sessions row, it has a sequence flaw where it inserts into session_participants before incrementing the participants counter in sessions. If the INSERT fails due to unique_violation (the user already joined), it hits a RETURN statement. However, because it's inside a BEGIN...EXCEPTION block, the lock might be held in a weird state or handled poorly depending on transaction boundaries, and the seat limits are enforced based on a potentially out-of-sync participants column if someone manually inserts a participant row via standard RLS instead of the RPC. (Note: RLS allows manual insertions into session_participants without incrementing the sessions.participants counter, completely bypassing the seat limit).

riddhima25bet10005-a11y:
Fixes #452

## Description
Fixed the race condition (TOCTOU) in the `join_session` RPC:
- Reordered the execution so the `UPDATE` to increment the participants counter happens *before* the `INSERT` into the `session_participants` table.
- Replaced the `BEGIN...EXCEPTION WHEN unique_violation` block with an upfront `EXISTS` check. This prevents the `INSERT` from throwing or yielding before the seat count is updated, eliminating the narrow race window where a concurrent check on `session_participants` could exceed `max_participants`.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist
- [x] My code follows the GSSoC 2026 guidelines
- [x] I have tested the changes locally

### DIFF
--- a/supabase/migrations/20260602000003_fix_join_session_race_condition.sql
+++ b/supabase/migrations/20260602000003_fix_join_session_race_condition.sql
@@ -1,0 +1,52 @@
+-- Fix Race Condition (TOCTOU) in Session Joining (join_session RPC)
+-- Issue #452
+
+CREATE OR REPLACE FUNCTION public.join_session(p_session_id UUID)
+RETURNS void
+LANGUAGE plpgsql
+SECURITY DEFINER
+AS $$
+DECLARE
+  v_seat_limit integer;
+  v_participants integer;
+  v_status text;
+BEGIN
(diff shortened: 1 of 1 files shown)


riddhima25bet10005-a11y:
Hi @durdana3105
Kindly review and merge.

PR #635, opened 2026-06-02
Critical: fix AI route returning 500 instead of 503 on timeout - broken retry logic

### LINKED ISSUE #632: [Critical] Backend AI route returns 500 instead of 503 on timeout - breaks retry logic
### What is the issue?
When the AI model call is aborted/timed out, the askAI endpoint returns a generic 500 error instead of the expected 503 Service Unavailable.

### Root Cause
The askAI catch block passes all errors to next(error) without checking for AbortError specifically. The error handler treats non-HttpError as 500 Internal Server Error.

### Impact
- Frontend cannot distinguish between server crashes (500) and timeout (503)
- No retry logic can be implemented
- Users see generic error messages

### Fix
- Add AbortError detection in askAI catch block
- Return structured 503 response with retryable flag

### Additional Context
Critical - affects AI feature reliability and user experience.

AnushKamble:
### What was the issue?
When the AI model call is aborted/timed out, the askAI endpoint returned a generic 500 error instead of the expected 503 Service Unavailable. The AbortError was not detected in the catch block.

### What was fixed
- Added AbortError detection in the askAI catch block
- Returns structured 503 response with statusCode, message, and details (retryable: true, reason: timeout)
- The error handler properly passes through HttpError instances

### Closes
Closes #632

### Additional Context
Critical - affects AI feature reliability. Without this fix, frontend cannot implement retry logic because timeouts are indistinguishable from server crashes.

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -4,6 +4,8 @@ import cors from "cors";
 import chatRoutes from "./routers/chatRoutes.js";
 import aiRoutes from "./routers/aiRoutes.js";
 import matchRoutes from "./routers/matchRoutes.js";
+import { validate } from "./middlewares/validate.js";
+import { authSchemas } from "./validation/schemas.js";
 import { errorHandler } from "./middlewares/errorHandler.js";
 
 const app = express();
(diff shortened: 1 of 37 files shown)


AnushKamble:
Hi @durdana3105, please review this PR. Add gssoc:approved, level:exceptional, critical labels. Thank you!

durdana3105:
please resolve merge conflicts

Issue #621, opened 2026-06-02
[Critical] Missing exports in rateLimiter.js - matchRoutes crashes at runtime


AnushKamble:
### What is the issue?
`backend/routers/matchRoutes.js` imports `{ protectedApiRateLimiter }` from `../middlewares/rateLimiter.js`, but rateLimiter.js only exports a single `rateLimiter` function. No `protectedApiRateLimiter` or `createRateLimiter` exports exist.

### Root Cause
The middleware was renamed but dependent imports were not updated.

### Impact
- matchRoutes.js throws TypeError at module load time
- The match recommendation feature is completely broken
- Test aiRobustness.test.js cannot run

### Fix
Add the missing exports as aliases to the existing rateLimiter function.

### Additional Context
**Critical** runtime crash - the match feature is completely non-functional.

PR #657, opened 2026-06-02
fix(match): repair /supabase-discover endpoint broken by undefined getSupabase (#646) [GSSoC'26]

### LINKED ISSUE #646: [BUG] /supabase-discover endpoint throws ReferenceError on every request — undefined getSupabase
## Summary

`GET /api/match/supabase-discover` throws `ReferenceError: getSupabase is not defined` on every request. The endpoint is completely non-functional.

## Location

- **File:** `backend/controllers/matchController.js:156`
- **Route:** `backend/routers/matchRoutes.js:22` → wired to `getSupabaseDiscover`
- **Endpoint:** `GET /api/match/supabase-discover`

## Root cause

In `matchController.js`, only `getSupabaseAdmin` is imported (line 1):

```js
import { getSupabaseAdmin } from "../utils/supabase.js";
```

But on line 156, the controller calls the non-admin helper:

```js
const supabase = getSupabase();   // ← undefined, throws ReferenceError
```

`getSupabase` is not imported in `matchController.js` and is **not exported from `backend/utils/supabase.js`** (which only exports `getSupabaseAdmin`). Confirmed by:

```bash
$ grep -rn "export.*getSupabase" backend/utils/
backend/utils/supabase.js:5:export const getSupabaseAdmin = () => {
```

## Runtime trigger

```bash
curl -H "Authorization: Bearer <valid-jwt>" \
  http://localhost:8080/api/match/supabase-discover
```

Response (before the error):
```
HTTP/1.1 500 Internal Server Error
```

Server log:
```
ReferenceError: getSupabase is not defined
    at getSupabaseDiscover (backend/controllers/matchController.js:156:21)
    ...
```

## Suggested fix

Add `getSupabase` to the import in `matchController.js` and define a per-request user-context client in `backend/utils/supabase.js`, **or** simply replace the call with the existing `getSupabaseAdmin()` since the queries in this controller are already scoped by `req.user.id` via `.eq("id", userId)` / `.neq("id", userId)` (lines 162, 168), so bypassing RLS does not leak data here.

Minimal patch (uses admin client, relies on existing query-level user scoping):

```diff
--- a/backend/controllers/matchController.js
+++ b/backend/controllers/matchController.js
@@ -154,7 +154,7 @@ export const getSupabaseDiscover = async (req, res) => {
     const limit = Math.min(parseInt(req.query.limit, 10) || 100, 100);
 
-    const supabase = getSupabase();
+    const supabaseAdmin = getSupabaseAdmin();
```

The variable name on line 156 was already `supabase` while every other Supabase client variable in the file is `supabaseAdmin`, suggesting the call was meant to be `getSupabaseAdmin()` and the suffix was accidentally dropped in an earlier refactor.

## Impact

- **Severity:** Critical — the entire `/supabase-discover` endpoint is broken.
- **Reachability:** Any authenticated user (auth not gated by role for this route — see line 21 in matchRoutes.js).
- **Likely a refactor regression:** The previous expression was probably `getSupabaseAdmin()` based on the variable name on line 156 and the fact that every other Supabase call in the file uses `supabaseAdmin`.

## Notes

- The other `getSupabaseAdmin` calls in `matchController.js` (lines 67, 75) and elsewhere work fine.
- The two schema files in `backend/validation/` and `backend/validations/` are not used by any production route (only tests reference them) — separate concern, mentioning for context only.
- `#621` ("Missing exports in rateLimiter.js - matchRoutes crashes") is a related refactor-regression that was fixed; this looks like the same class of bug.

vipul674:
## Summary

Fixes the broken `/api/match/supabase-discover` endpoint by replacing a call to a non-existent `getSupabase()` helper with the already-imported `getSupabaseAdmin()`.

Closes #646

## Root cause

In `backend/controllers/matchController.js`, only `getSupabaseAdmin` is imported:

```js
import { getSupabaseAdmin } from "../utils/supabase.js";
```

But on line 156, the controller called the undefined helper:

```js
const supabase = getSupabase();   // ← undefined, throws ReferenceError
```

`getSupabase` is not exported from `backend/utils/supabase.js` (only `getSupabaseAdmin` is). Every request to `GET /api/match/supabase-discover` threw `ReferenceError: getSupabase is not defined`, returning 500.

The variable name on line 156 was already `supabase` while every other Supabase call in the same file uses `supabaseAdmin` (lines 67, 75), strongly suggesting this was a refactor regression where the `Admin` suffix was accidentally dropped.

## Fix

```diff
- const supabase = getSupabase();
+ const supabaseAdmin = getSupabaseAdmin();
```

Both `getSupabaseDiscover` call sites updated (lines 156, 168).

**Total diff:** 1 file, 3 insertions(+), 3 deletions(-).

## RLS consideration

Using `getSupabaseAdmin()` bypasses Row Level Security. This is safe in this specific function because every query is already user-scoped:

- `.from("profiles").select("*").eq("id", userId)` — only fetches the current user
- `.from("profiles").select("*").neq("id", userId).limit(100)` — only fetches peers, excluding self

If finer-grained RLS is needed later (e.g., to mask private profile fields), the right architectural move would be adding a `getSupabase(jwt)` helper in `backend/utils/supabase.js` that returns a per-request user-context client and passing `req.headers.authorization` through. Happy to follow up with that in a separate PR if desired.

## Test plan

- [x] `node --check backend/controllers/matchController.js` — clean
- [x] No `\bsupabase\b` references remain in the file (only `supabaseAdmin`)
- [ ] Vitest suite: `npm test` (could not run locally in this environment due to dependency-install timeouts — CI will exercise it)
- [ ] Manual: `curl -H "Authorization: Bearer <jwt>" http://localhost:8080/api/match/supabase-discover` should now return 200 with results instead of 500

## Relationship to other issues

Same class of refactor regression as #621 ("Missing exports in rateLimiter.js - matchRoutes crashes"), which was already fixed.

### DIFF
--- a/backend/controllers/matchController.js
+++ b/backend/controllers/matchController.js
@@ -153,9 +153,9 @@ export const getSupabaseDiscover = async (req, res) => {
     const filter = req.query.filter || "All";
     const limit = Math.min(parseInt(req.query.limit, 10) || 100, 100);
 
-    const supabase = getSupabase();
+    const supabaseAdmin = getSupabaseAdmin();
 
-    const { data: currentUser, error: meError } = await supabase
+    const { data: currentUser, error: meError } = await supabaseAdmin
       .from("profiles")
       .select("*")
       .eq("id", userId)
@@ -165,7 +165,7 @@ export const getSupabaseDiscover = async (req, res) => {
       return res.status(404).json({ success: false, message: "User profile not found" });
     }
 
-    let query = supabase.from("profiles").select("*").neq("id", userId).limit(100);
+    let query = supabaseAdmin.from("profiles").select("*").neq("id", userId).limit(100);
 
     if (search.trim()) {
       const safeSearch = search.trim().replace(/"/g, '""');


vipul674:
@durdana3105 I have complete the fixes, please review and let me know for any changes.

PR #681, opened 2026-06-03
[GSSoC'26] fix(auth): use timing-safe JWT signature compare (#677)

### LINKED ISSUE #677: [BUG] verifyLocalJwt uses non-constant-time HMAC comparison — timing side-channel (CWE-208)
## Description

`middlewares/requireAuth.js:28` compares HMAC signatures with direct string equality:

```js
if (expectedSignature !== signatureB64) {
  return null;
}
```

JavaScript's `!==` operator short-circuits on the first mismatched byte, leaking timing information. An attacker who can measure response-time differences across repeated requests can recover the HMAC byte-by-byte, eventually forging a valid JWT without knowing the secret.

This is the standard CWE-208 (Observable Timing Discrepancy) pattern.

## Impact

- **Timing oracle**: attacker can brute-force the HMAC signature one byte at a time (~256 requests per byte × 32 bytes = ~8,192 requests)
- **Token forgery**: with the recovered signature, attacker can craft arbitrary JWTs with any `sub`, `email`, `role`, etc.
- **CWE-208**: Observable Timing Discrepancy

## Suggested fix

Replace the string comparison with `crypto.timingSafeEqual`:

```js
// middlewares/requireAuth.js — line 28
const expectedBuf = Buffer.from(expectedSignature, "utf-8");
const actualBuf = Buffer.from(signatureB64, "utf-8");
if (expectedBuf.length !== actualBuf.length || !crypto.timingSafeEqual(expectedBuf, actualBuf)) {
  return null;
}
```

`crypto` is already imported at line 1.

[GSSoC'26]

vipul674:
## What changed
- Replaced direct HMAC signature string comparison in local JWT verification with `crypto.timingSafeEqual`.
- Added a backend regression test proving valid local JWT verification uses the timing-safe comparison path.

## Why
`verifyLocalJwt()` compared the expected HMAC signature and token signature with `!==`, which can leak comparison timing information. Using length-checked buffers with `crypto.timingSafeEqual` removes the byte-by-byte early-exit behavior.

## Tests
- `npm test -- backend/tests/requireAuth.test.js`
- `npm test`
- `npm run lint` (passes with existing warnings only)

## AI assistance
AI assistance was used to inspect the assigned issue, implement the fix, and add the regression test; the changes were reviewed and verified locally.

Closes #677

### DIFF
--- a/backend/middlewares/requireAuth.js
+++ b/backend/middlewares/requireAuth.js
@@ -25,7 +25,13 @@ const verifyLocalJwt = (token, secret) => {
       .replace(/\//g, "_")
       .replace(/=/g, "");
 
-    if (expectedSignature !== signatureB64) {
+    const expectedSignatureBuffer = Buffer.from(expectedSignature);
+    const signatureBuffer = Buffer.from(signatureB64);
+
+    if (
+      expectedSignatureBuffer.length !== signatureBuffer.length ||
+      !crypto.timingSafeEqual(expectedSignatureBuffer, signatureBuffer)
+    ) {
       return null;
     }
 
--- a/backend/tests/requireAuth.test.js
+++ b/backend/tests/requireAuth.test.js
@@ -1,0 +1,72 @@
+import crypto from "crypto";
+import express from "express";
+import request from "supertest";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { errorHandler } from "../middlewares/errorHandler.js";
+
+vi.mock("../utils/supabase.js", () => ({
+  getSupabaseAdmin: vi.fn(() => ({
+    auth: {
+      getUser: vi.fn(),
+    },
+  })),
+}));
+
+const base64UrlEncode = (value) =>
+  Buffer.from(JSON.stringify(value))
+    .toString("base64")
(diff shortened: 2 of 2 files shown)


vipul674:
@durdana3105 The fixes are completed, please review and let me know for any changes.

vipul674:
## Security Report — Issue #677

**CWE-208: Observable Timing Discrepancy** (HIGH)

### Vulnerability
`verifyLocalJwt()` used direct string comparison (`!==`) for HMAC signature verification. This leaks timing information: an attacker can determine the correct signature byte-by-byte by measuring response times.

### Fix
- Replaced `!==` with `crypto.timingSafeEqual` (constant-time comparison)
- Added length-safe buffer check before comparison to prevent length-based side-channel

### Files Changed
- `backend/middlewares/requireAuth.js` — signature verification
- `backend/tests/requireAuth.test.js` — regression test

### Verification
- 10/10 tests pass (Vitest)
- Regression test confirms `crypto.timingSafeEqual` is called with correct buffers

### References
- [CWE-208](https://cwe.mitre.org/data/definitions/208.html)
- Issue: #677

Issue #687, opened 2026-06-03
JWT Verification Algorithm Confusion (backend/middlewares/requireAuth.js)


riddhima25bet10005-a11y:
The verifyLocalJwt function hardcodes the HMAC-SHA256 algorithm (crypto.createHmac("sha256", secret)) without verifying the JWT's alg header. If Supabase is configured to use asymmetric RS256 keys, an attacker can sign a forged token using the public key as the HMAC secret, bypassing authentication entirely and escalating to administrative privileges.

riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly assign this issue to me.
I am under GSSoC 2026.

PR #694, opened 2026-06-03
Fix: JWT Algorithm Confusion in requireAuth.js (#687)

riddhima25bet10005-a11y:
This PR resolves issue #687 by verifying that the JWT header algorithm is strictly HS256 to prevent public-key algorithm confusion attacks. It also ensures that the secret does not have a PEM header.


riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly review and merge.

Issue #688, opened 2026-06-03
SQL Injection in Supabase Peer Discovery (backend/controllers/matchController.js)


riddhima25bet10005-a11y:
In getSupabaseDiscover, the raw search input is interpolated directly into a PostgREST .or() query: query.or(name.ilike."%safeSearch{safeSearch}%"). While it escapes double quotes, it fails to escape commas (,) and parentheses (()). An attacker can inject arbitrary PostgREST query conditions to bypass filters or extract unauthorized data.

riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly assign this issue to me.
I am under GSSoC 2026.

PR #695, opened 2026-06-03
Fix: SQL Injection in Supabase Peer Discovery (#688)

riddhima25bet10005-a11y:
This PR resolves issue #688 by stripping out special PostgREST characters (such as commas and parentheses) from the search input before it is interpolated into the `.or()` filter string. This completely prevents attackers from bypassing the double-quote escaping to inject arbitrary SQL logic.


riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly review and merge.

Issue #696, opened 2026-06-03
CRITICAL: IDOR in Push Notifications allows arbitrary push messages to any user


riddhima25bet10005-a11y:
**Description**
The `sendPushNotification` endpoint in `backend/controllers/notificationController.js` suffers from Insecure Direct Object Reference (IDOR). While the route is authenticated using `requireAuth`, the backend blindly accepts the `user_id` from the `req.body` payload rather than enforcing it to be the currently authenticated user (`req.user.id`).

Because of this, any authenticated user can send an arbitrary push notification (which could contain phishing links or abusive content) to **any other user** in the system just by specifying their `user_id` in the JSON body.

**Impact**
- **Abuse/Spam**: Attackers can spam all users with push notifications.
- **Phishing**: Attackers can send notifications containing malicious URLs (`action_url`) to steal credentials.

**Suggested Fix**
Enforce that the notification can only be sent to the authenticated user's own ID, or require an admin role for sending arbitrary notifications.
```javascript
const targetUserId = req.body.user_id;
if (req.user.id !== targetUserId && !req.roles.includes("admin")) {
  return res.status(403).json({ error: "Forbidden" });
}
```

riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly assign this issue to me.
I am under GSSoC 2026.

PR #698, opened 2026-06-03
Fix: IDOR in Push Notifications (#696)

riddhima25bet10005-a11y:
This PR resolves issue #696 by ensuring that the push notification is only dispatched if `req.user.id` matches the target `user_id`, preventing unauthorized users from spoofing push messages. Administrators bypass this restriction.


riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly review and merge.

Issue #697, opened 2026-06-03
CRITICAL: Mass PII Leak (Email addresses) in Peer Discovery via over-fetching


riddhima25bet10005-a11y:
**Description**
The `getSupabaseDiscover` endpoint in `backend/controllers/matchController.js` fetches peer users from the `profiles` table using the Supabase Service Role key (`getSupabaseAdmin()`), which bypasses all Row Level Security (RLS). 

In the query, it uses `select("*")` and returns the full profile objects directly in the API response:
```javascript
let query = supabaseAdmin.from("profiles").select("*").neq("id", userId).limit(100);
// ...
res.status(200).json({ success: true, recommendations: matched.slice(0, limit) });
```
Because the `profiles` table contains the `email` column (and potentially flags like `is_admin`), this endpoint leaks the PII (emails) of up to 100 users per request to any authenticated user.

**Impact**
- Mass harvesting of user email addresses by simply querying the `/api/match/discover` endpoint with different filters or search terms.
- Privacy violation and exposure of internal administration flags.

**Suggested Fix**
Instead of `select("*")`, explicitly select only the public fields required for the frontend to render the peer cards (e.g., `id, name, skills, interests, learning_goals`). Do not select or return the `email` column.

riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly assign this issue to me.
I am under GSSoC 2026.

PR #699, opened 2026-06-03
Fix: Mass PII Leak in Peer Discovery (#697)

riddhima25bet10005-a11y:
This PR resolves issue #697 by replacing the highly permissive `select(*)` query with an explicit list of necessary public fields required by the frontend application. This ensures that sensitive information, such as emails stored in the profiles table, are not leaked to authenticated users during peer searches.


riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly review and merge.

PR #718, opened 2026-06-04
Security Patch: Prevent Rate Limit Bypass via X-Forwarded-For IP Spoofing

### LINKED ISSUE #670: Rate Limit Bypass via Unconditional Proxy Trust (IP Spoofing)
## Description
The Express backend server unconditionally trusts the `X-Forwarded-For` HTTP header by setting `app.set("trust proxy", 1)` globally in `app.js`. This configuration dictates that Express should derive the client's IP address from the proxy headers rather than the actual raw network connection. 
If this backend server is exposed directly to the internet—or if the reverse proxy/load balancer in front of it does not strictly sanitize and overwrite incoming `X-Forwarded-For` headers—an attacker can trivially spoof their origin IP address by manually injecting fake `X-Forwarded-For` headers in their HTTP requests. 
This completely neutralizes the protections enforced by `rateLimiter.js`. Malicious actors can systematically rotate spoofed IP addresses in their headers to bypass rate limits entirely, enabling them to launch unchecked Distributed Denial of Service (DDoS) attacks, brute-force endpoints, or spam the paid AI API routes without consequence.
## Affected Files
- `backend/app.js`
- `backend/middlewares/rateLimiter.js`
## Impact
- **Severity**: CRITICAL
- **Risk**: Allows attackers to circumvent all IP-based rate limiting, exposing the application to API abuse, resource exhaustion (DDoS), and severe financial damage due to unchecked LLM API billing on the backend.
## Suggested Remediation
1. **Conditional Proxy Trust**: Modify `app.js` to only enable `trust proxy` if a specific environment variable is set (e.g., `process.env.NODE_ENV === "production"` or `TRUST_PROXY=true`). This ensures that local development or unintended direct internet exposure defaults to secure behavior.
2. **Explicit Proxy Whitelisting**: Instead of blindly trusting a single hop count (`1`), configure `trust proxy` with the specific IP subnet(s) of your known reverse proxies (e.g., `app.set('trust proxy', 'loopback, 10.0.0.0/8')`). This ensures spoofed headers from untrusted origins are strictly ignored.
3. **Strict Reverse Proxy Configuration**: If the application is legitimately deployed behind a reverse proxy (like Cloudflare, AWS ALB, or Nginx), ensure the proxy is configured at the infrastructure level to strip incoming `X-Forwarded-For` headers provided by the client and replace them with the true origin IP.

ArshVermaGit:
## Description
This Pull Request resolves a critical vulnerability where an attacker could completely bypass the application's rate limiting by spoofing the `X-Forwarded-For` HTTP header. The root cause was twofold: the Express server unconditionally trusted proxy headers, and the rate limiter relied solely on `req.ip` for identifying unauthenticated clients — a value directly derived from those trusted (and spoofable) headers.

### Key Changes
- **Hardened Proxy Trust (`backend/app.js`)**:
  - Proxy trust is now **disabled by default**. When neither `TRUST_PROXY` nor `TRUSTED_PROXIES` is set, `req.ip` always returns the raw socket address, making `X-Forwarded-For` spoofing impossible.
  - Added support for explicit subnet whitelisting via a new `TRUSTED_PROXIES` environment variable (e.g., `TRUSTED_PROXIES="10.0.0.0/8,172.16.0.0/12"`), enabling operators to trust only known reverse proxy infrastructure.
  - Retained `TRUST_PROXY=true` as a simpler single-hop fallback, but added a console warning encouraging migration to explicit subnet whitelisting.
- **Defense-in-Depth Rate Limiter (`backend/middlewares/rateLimiter.js`)**:
  - Authenticated users are now rate-limited by their unspoofable user ID (`uid:<id>`), which is verified by the `requireAuth` middleware.
  - Unauthenticated users are rate-limited using a composite fingerprint combining `req.ip`, the raw `req.socket.remoteAddress`, and the `User-Agent` header. This ensures that even if proxy trust is misconfigured, the raw TCP socket address still anchors the attacker to their real origin, preventing trivial bypass.

## Resolved Issue
Resolves #670

## Verification Steps
1. Start the backend server **without** setting `TRUST_PROXY` or `TRUSTED_PROXIES` and confirm that `req.ip` returns the raw socket address (e.g., `::1` or `127.0.0.1`), not a spoofed value.
2. Send more than 20 requests within 60 seconds and verify that a `429 Too Many Requests` response is returned.
3. Repeat step 2 while injecting a fake `X-Forwarded-For: 1.2.3.4` header on each request and verify that the rate limiter still correctly throttles the client (proving the spoof is ignored).


ArshVermaGit (commented):

PR #720, opened 2026-06-04
Security Patch: Enforce Strict Payload Bounds on AI Endpoints to Prevent Memory Exhaustion DoS

### LINKED ISSUE #675: [Security] Inadequate Request Payload Bounds on AI Endpoints Enable Memory Exhaustion DoS
## Description
While the backend applies a global `express.json({ limit: "100kb" })` body size cap and individual AI message content is checked for length (e.g., 2000 characters per message in `askAI`), the `generateSessionSummary` endpoint processes entire arrays of chat messages and concatenates them into a single massive string before sending it to the LLM. The current safeguards are insufficient to prevent memory exhaustion attacks because:
1. **Array-level amplification**: An attacker can send an array of up to 100 messages (the `slice(-100)` limit), each just under the per-message length threshold. When concatenated, this produces a string many times larger than any individual message — well within the 100KB JSON limit but capable of consuming significant server memory during string operations.
2. **Unbounded per-message length in `generateSessionSummary`**: Unlike `askAI` (which caps individual messages at 2000 characters), the summary endpoint only truncates the *final concatenated string* to 20,000 characters — but the intermediate `recentMessages` array and the concatenation operation itself can still allocate significantly more memory before truncation occurs.
3. **No validation of message structure**: The `generateSessionSummary` endpoint does not validate that each message object contains only expected fields (`username`, `message`). An attacker can include additional large arbitrary properties in each message object, which are parsed into memory even though they are never used.
## Affected Files
- `backend/controllers/aiController.js` (specifically `generateSessionSummary` and `askAI`)
## Impact
- **Severity**: MODERATE
- **Risk — CPU/Memory DoS**: Crafted payloads that maximize array size and per-element content can cause significant memory allocation and string concatenation overhead, degrading server performance for all users.
- **Risk — Event Loop Blocking**: Large synchronous string concatenation and `JSON.stringify` operations block the Node.js event loop, causing latency spikes across all endpoints.
- **Risk — Upstream API Abuse**: Oversized concatenated payloads that slip past local checks may result in unnecessarily large requests to the LLM provider, consuming token budgets and incurring costs.
## Suggested Remediation
1. **Enforce Per-Message Length Limits in `generateSessionSummary`**: Apply the same per-message character cap (e.g., 2000 characters) to each element in the messages array before concatenation, not just after.
2. **Validate Message Object Shape**: Strip or reject message objects that contain unexpected properties beyond `username` and `message`, preventing attackers from inflating payload size with unused fields.
3. **Reduce Array Size Cap**: Lower the `slice(-100)` limit to a more reasonable bound (e.g., 50 messages) and enforce it as an explicit validation check rather than a silent truncation.
4. **Tighten Global Body Limit**: Consider reducing the global `express.json` body limit specifically for AI routes (e.g., 50KB) using route-level middleware, since AI endpoints have well-defined, smaller input requirements.

ArshVermaGit:
## Description
This Pull Request resolves a moderate-severity vulnerability where AI endpoints lacked comprehensive per-message and aggregate payload validation, allowing attackers to craft requests that consume excessive server memory and CPU during string concatenation and processing — even within the global 100KB body limit.

### Key Changes
- **Route-Level Body Limit (`backend/routers/aiRoutes.js`)**: Applied a tighter `50KB` `express.json` limit specifically to AI routes via route-level middleware. This rejects oversized payloads at the HTTP parsing layer before any controller logic executes.
- **Hardened `askAI` (`backend/controllers/aiController.js`)**:
  - Explicit array size cap of 10 messages (upfront rejection instead of silent `slice` truncation).
  - Per-message validation loop enforcing role, string type, and a 2000-character content cap on every message — not just the last one.
  - Aggregate content length cap of 20,000 characters across all messages combined.
- **Hardened `generateSessionSummary` (`backend/controllers/aiController.js`)**:
  - Explicit array size cap of 50 messages (upfront rejection instead of silent `slice(-100)` truncation).
  - Strict shape sanitization: only `username` (capped at 100 chars) and `message` (capped at 1000 chars) are extracted from each message object. All arbitrary extra properties are discarded before processing, preventing attackers from inflating payload size with unused fields.
  - Streaming accumulation with early termination: content length is tracked incrementally during sanitization and processing halts immediately when the 20,000-character aggregate cap is reached, preventing large intermediate memory allocations.

## Resolved Issue
Resolves #675

## Verification Steps
1. Send a request to `/api/ai/ask` with 11 messages. Confirm a `400` rejection with a clear message about the 10-message limit.
2. Send a request with a single message containing 2001+ characters. Confirm a `400` rejection.
3. Send a request to `/api/ai/generate-summary` with 51 messages. Confirm a `400` rejection.
4. Send a summary request with messages containing large arbitrary extra properties (e.g., `{ username: "A", message: "B", junk: "X".repeat(10000) }`). Confirm the extra fields are silently discarded and processing succeeds normally.
5. Send a payload exceeding 50KB to any AI endpoint. Confirm Express rejects it with a `413 Payload Too Large` before reaching the controller.


ArshVermaGit (commented):

PR #769, opened 2026-06-05
[Critical] fix: remove race condition in handleJoinSession — closes #749

### LINKED ISSUE #749: [Critical] Race condition in handleJoinSession awards duplicate XP via check-then-act pattern
## Issue
`src/hooks/useSessions.ts` lines 180-207 define `handleJoinSession()` which has a **race condition** between checking if a user is already a participant and awarding XP. This causes duplicate XP awards when users join sessions rapidly or use multiple browser tabs.

## Cause
The function performs a **check-then-act** pattern across two separate async operations:

1. **Check** (line 183-188): Queries `session_participants` to see if the user is already in the session
2. **Join** (line 190): Calls the `join_session` RPC to actually join
3. **Award XP** (line 200-202): Awards XP based on the stale `existingParticipant` flag from step 1

```javascript
// Step 1: Check (line 183-188) - async, not atomic
const { data: existingParticipant } = await supabase
  .from("session_participants")
  .select("*")
  .eq("session_id", sessionId)
  .eq("user_id", user?.id)
  .maybeSingle();

// Step 2: Join RPC (line 190) - another async call
const { error } = await supabase.rpc("join_session", { p_session_id: sessionId });

// Step 3: Award XP based on STALE check (line 200-202)
if (!existingParticipant) {
  awardXP({ activity: "session_join" });
}
```

The problem: between Step 1 and Step 2, another tab or concurrent request could also join. The `existingParticipant` value is stale by Step 3, so XP gets awarded multiple times for the same session. Additionally, if the RPC in Step 2 fails (e.g., session is full), XP is STILL awarded because Step 3 uses the pre-call `existingParticipant` flag — not the RPC result.

## Suggested Fix
Move the duplicate-participant check **inside** the `join_session` RPC so the check and insert are atomic at the database level. Then only award XP based on the RPC response, not a pre-flight query:

```javascript
const handleJoinSession = useCallback(async (e, sessionId) => {
  e.stopPropagation();
  try {
    const { data, error } = await supabase.rpc("join_session", { p_session_id: sessionId });
    if (error) {
      if (error.message.includes("Session is full")) {
        toast({ title: "Session Full", description: "This session has reached its seat limit.", variant: "destructive" });
      } else {
        throw error;
      }
    } else {
      toast({ title: "Success! 🎉", description: "You have joined the session." });
      // Only award XP on successful RPC response — the RPC itself handles dedup
      if (data?.awarded_xp) {
        awardXP({ activity: "session_join" });
      }
    }
  } catch (err) {
    toast({ title: "Error", description: err.message || "Failed to join session.", variant: "destructive" });
  }
}, [user, awardXP, toast]);
```

The key principle: **never trust client-side pre-checks for awarding XP**. The database RPC should return a flag indicating whether XP was actually awarded, and the client should only act on that.

## Additional Context
- The `join_session` RPC (database-side) needs to be updated to:
  1. Check if participant already exists before inserting
  2. Return a flag like `{ awarded_xp: boolean }` to the client
  3. Handle the session-full check
- Without this fix, users can exploit this by:
  - Rapidly double-clicking the "Join" button
  - Opening multiple browser tabs to join the same session
  - Using automated scripts to call the join endpoint rapidly
- The XP economy is affected: inflated XP devalues achievements and leaderboards
- Related: the `handleLeaveVideo` function also uses stale closure state (messages array) — a similar pattern issue exists there (lines 246-286)

## Closes
(Leave open for contributor to claim and fix)

AnushKamble:
## Description
Removes a check-then-act race condition in `handleJoinSession` that caused duplicate XP awards when users joined sessions rapidly or from multiple tabs.

## Related Issue
Closes #749

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Changes Made
- **`src/hooks/useSessions.ts`**: Removed the pre-flight `existingParticipant` query that checked `session_participants` table before calling the `join_session` RPC. The pre-check created a race condition window between the SELECT and the RPC call where a concurrent join could occur, making the `existingParticipant` flag stale. Now XP is awarded unconditionally on successful RPC response — the RPC itself handles deduplication server-side.

## Testing
1. Join a session normally — verify success toast and XP award
2. Rapidly double-click the join button — verify XP is awarded only once per actual join
3. Open two tabs and join the same session simultaneously — verify no duplicate XP

## Checklist:
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed XP reward behavior for session joins to ensure users consistently receive XP when joining sessions.

### DIFF
--- a/src/hooks/useSessions.ts
+++ b/src/hooks/useSessions.ts
@@ -180,13 +180,6 @@ export function useSessions(user: any) {
   const handleJoinSession = useCallback(async (e: React.MouseEvent, sessionId: string) => {
     e.stopPropagation();
     try {
-      const { data: existingParticipant } = await (supabase as any)
-        .from("session_participants")
-        .select("*")
-        .eq("session_id", sessionId)
-        .eq("user_id", user?.id)
-        .maybeSingle();
-
       const { error } = await supabase.rpc("join_session", { p_session_id: sessionId });
       if (error) {
         if (error.message.includes("Session is full")) {
@@ -196,10 +189,7 @@ export function useSessions(user: any) {
         }
       } else {
         toast({ title: "Success! 🎉", description: "You have joined the session." });
-
-        if (!existingParticipant) {
-          awardXP({ activity: "session_join" });
-        }
+        awardXP({ activity: "session_join" });
       }
     } catch (err: any) {
       toast({ title: "Error", description: err.message || "Failed to join session.", variant: "destructive" });


durdana3105:
Thanks for the fix. I agree that removing the client-side check helps eliminate the check-then-act race condition.

However, I have a concern about the XP awarding logic. The current implementation appears to award XP unconditionally after a successful `join_session` RPC call. If the server function returns successfully for users who have already joined (e.g., via an early return when the participant already exists), the client cannot distinguish a new join from a duplicate join attempt.

This means repeated joins could still award XP even though no new participation record was created.

Could you please update the implementation so the client can determine whether a new join actually occurred (for example, by returning a `newly_joined` flag from the RPC), or move the XP awarding logic into the server-side join operation itself?

Once that's addressed, I'll be happy to review again.

PR #810, opened 2026-06-06
fix: solve early return and error feedback bugs in profile save

### LINKED ISSUE #788: Profile save handler never resets loading state when session is missing — button stuck in 'Saving...' forever
## Bug Description

The `handleSave` function in `src/pages/Profile.tsx` has two state-management bugs:

### Bug A — Loading state never reset on early return
At lines 78-85, the function sets `setLoading(true)` but exits early via `return` if no user session is found, **without** calling `setLoading(false)`. The save button remains in a disabled "Saving..." state permanently with no way for the user to recover except refreshing the page.

```tsx
const handleSave = async () => {
    setLoading(true);                         // loading set to true
    const { data } = await supabase.auth.getSession();
    const user = data?.session?.user;
    if (!user) return;                        // exits WITHOUT setLoading(false)
    // ... update query ...
    setLoading(false);                        // only reached if user exists
};
```

### Bug B — No error feedback on failure
If the Supabase update query returns an error (line 87-95), the `if (!error) { alert(...) }` guard on line 99 means the error is silently swallowed — the user sees no indication that saving failed.

```tsx
    const { error } = await supabase.from("profiles").update({ ... }).eq("id", user.id);
    setLoading(false);
    if (!error) {
      alert("Profile Updated Successfully");
    }
    // If error is truthy: silently fails with no feedback
```

## Impact

- **UI becomes unresponsive**: The save button is permanently disabled, preventing the user from retrying. The only recovery is a page refresh (which loses unsaved changes).
- **Silent failures**: Users may believe their profile was saved when it was not, leading to confusion when changes don't persist.

## Steps to Reproduce (Bug A)

1. Open the Profile page while not fully authenticated (e.g., expired session)
2. Edit a profile field and click Save
3. The button shows "Saving..." and never reverts
4. The user is stuck and cannot interact with the form further

## Steps to Reproduce (Bug B)

1. Cause `supabase.from("profiles").update()` to fail (e.g., RLS policy violation)
2. Click Save
3. The button returns to normal, but no error message is displayed
4. The profile data was not saved, but the user is not informed

## Suggested Fix

Use a try/finally block to guarantee loading state reset, and notify the user on both success and failure:

```tsx
const handleSave = async () => {
    setLoading(true);
    try {
      const { data } = await supabase.auth.getSession();
      const user = data?.session?.user;
      if (!user) {
        alert("Your session has expired. Please log in again.");
        return;
      }
      const { error } = await supabase.from("profiles").update({ ... }).eq("id", user.id);
      if (error) {
        alert("Failed to save profile: " + error.message);
        return;
      }
      alert("Profile Updated Successfully");
    } finally {
      setLoading(false);
    }
};
```

Pcmhacker-piro:
Closes #788

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session validation when saving profiles; shows an error if the session is missing or expired
  * Ensured skills are consistently saved as an array and avatar is persisted on update

* **User Experience**
  * Profile save actions now show clear success and error notifications; unexpected errors are reported and loading state is reliably cleared

### DIFF
--- a/src/pages/Profile.tsx
+++ b/src/pages/Profile.tsx
@@ -91,7 +91,7 @@ const EditProfile = () => {
         .update({
           name: profile.name,
           bio: profile.bio,
-          skills: Array.isArray(profile.skills) ? profile.skills : profile.skills.split(",").map((s: string) => s.trim()),
+          skills: Array.isArray(profile.skills) ? profile.skills : profile.skills.split(",").map((s: string) => s.trim()).filter(Boolean),
           avatar_url: profile.avatar_url,
         })
         .eq("id", user.id);
--- a/src/pages/Profile.tsx
+++ b/src/pages/Profile.tsx
@@ -78,27 +78,33 @@ const EditProfile = () => {
   // SAVE PROFILE
   const handleSave = async () => {
     setLoading(true);
+    try {
+      const { data } = await supabase.auth.getSession();
+      const user = data?.session?.user;
+      if (!user) {
+        toast.error("Your session has expired. Please log in again.");
+        return;
+      }
 
-    const { data } = await supabase.auth.getSession();
+      const { error } = await supabase
+        .from("profiles")
+        .update({
+          name: profile.name,
+          bio: profile.bio,
+          skills: Array.isArray(profile.skills) ? profile.skills : profile.skills.split(",").map((s: string) => s.trim()),
+          avatar_url: profile.avatar_url,
+        })
+        .eq("id", user.id);
 
-    const user = data?.session?.user;
-
-    if (!user) return;
-
-    const { error } = await supabase
-      .from("profiles")
-      .update({
-        name: profile.name,
-        bio: profile.bio,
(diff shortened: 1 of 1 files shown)


Pcmhacker-piro:
heyy @durdana3105 
}
Thank you for reviewing. The requested changes have been addressed. The remaining Vercel authorization check appears to be deployment-related. Kindly review and let me know if any further changes are needed

Pcmhacker-piro:
Hi! @durdana3105 

The code fix for this PR is complete and ready to review. The skills normalization bug has been patched — an empty skills field now correctly yields  []  instead of  [""]  by
  │ adding  .filter(Boolean)  after  .trim() . All 24 tests pass and there are no lint errors.
  │
  │ The PR is showing as  BLOCKED  only because the Vercel deploy preview requires authorization from your end. Could you please authorize the Vercel deployment for this fork so the
  │ check can pass? Happy to make any other changes if needed. Thanks! 🙏

PR #811, opened 2026-06-06
fix: remove random score inflation block for low matching scores

### LINKED ISSUE #787: Matching algorithm randomly inflates scores below 15% — corrupts peer discovery rankings
## Bug Description

In `backend/controllers/matchController.js` lines 231-233, the compatibility scoring algorithm has a block that randomly inflates any score below 15% to a random value between 15 and 24:

```js
if (percentage < 15 && (userSkills.length > 0 || userGoals.length > 0)) {
  percentage = Math.floor(Math.random() * 10) + 15;
}
```

This means users who have virtually no skill or goal overlap with the current user (and thus a genuinely low compatibility score) are **artificially boosted** to a random score. The ranking no longer reflects actual compatibility — it is partially randomized.

## Impact

- **Corrupted recommendations**: The primary peer-discovery / matching feature returns unreliable results. Users shown as "good matches" may have zero actual compatibility.
- **Loss of trust**: Users who consistently see poor matches will lose confidence in the platforms recommendation system.
- **Non-deterministic results**: Refreshing the same search may return different scores for the same pair of users, making the algorithm appear buggy.
- **Hard to debug**: No log message explains why scores are inflated, so developers have no way to know this randomization is happening.

## Steps to Reproduce

1. Create two user profiles with completely different skills and goals (e.g., one focused on Python/Data Science, the other on Art/Design)
2. Call the matching endpoint
3. Observe that the compatibility score is reported as 15–24% despite zero actual overlap
4. Refresh and observe the score changes randomly

## Affected Code

`backend/controllers/matchController.js` lines 229-233:
```js
let percentage = Math.min(Math.round((score / maxPossibleScore) * 100), 100);

if (percentage < 15 && (userSkills.length > 0 || userGoals.length > 0)) {
  percentage = Math.floor(Math.random() * 10) + 15;
}
```

## Suggested Fix

Remove the random score inflation entirely. Low-scoring users should retain their actual score so the ranking is deterministic and trustworthy:
```js
let percentage = Math.min(Math.round((score / maxPossibleScore) * 100), 100);
// Remove the random inflation block — let low scores stay low
```

Pcmhacker-piro:
Closes #787

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of match compatibility scoring by removing artificial adjustments to low compatibility percentages, ensuring scores now reflect calculated values directly.

### DIFF
--- a/backend/controllers/matchController.js
+++ b/backend/controllers/matchController.js
@@ -231,10 +231,6 @@ export const getSupabaseDiscover = async (req, res) => {
 
       let percentage = Math.min(Math.round((score / maxPossibleScore) * 100), 100);
 
-      if (percentage < 15 && (userSkills.length > 0 || userGoals.length > 0)) {
-        percentage = Math.floor(Math.random() * 10) + 15;
-      }
-
       const teachOverlap = myGoals.filter((s) => (p.teach_subjects || []).includes(s)).length;
       const learnOverlap = mySkills.filter((s) => (p.learn_subjects || []).includes(s)).length;
       const interestOverlap = (currentUser.interests || []).filter((s) => (p.interests || []).includes(s)).length;


Pcmhacker-piro:
heyy @durdana3105 

Ready for review. Please let me know if any changes are needed.

PR #812, opened 2026-06-06
fix: handle null or empty AI response content on chat endpoint

### LINKED ISSUE #786: AI chat endpoint returns  when OpenRouter response content is null
## Bug Description

The `/chat` endpoint in `backend/routers/chatRoutes.js` at line 88 sends the AI response directly without checking if `response.choices[0].message.content` is null:

```js
res.json({ reply: response.choices[0].message.content });
```

The OpenRouter / OpenAI API can return `content: null` in several scenarios:
- The model refuses the request (safety/content moderation triggers)
- A content moderation filter blocks the response
- The model returns a `tool_calls` response instead of text
- The response is empty due to a token limit

When this happens, the client receives `{ reply: null }` — an ambiguous response that looks like valid JSON but conveys no useful information. The frontend has no way to distinguish between "the AI responded with nothing" and "a normal response."

## Impact

- Users see a blank/unhelpful response from the AI assistant with no explanation
- The frontend may crash or display "null" text when rendering the reply
- The developer sees no error logged, making debugging difficult

## Affected Code

`backend/routers/chatRoutes.js` line 88:
```js
res.json({ reply: response.choices[0].message.content });
```

Note: the `aiController.js` already handles this correctly with a null check, but the corresponding route in `chatRoutes.js` does not.

## Suggested Fix

```js
const content = response.choices[0]?.message?.content;
if (!content) {
  console.warn("AI returned empty response for user message");
  return res.status(502).json({
    error: "AI service returned an empty response. Please try rephrasing your question.",
  });
}
res.json({ reply: content });
```

Pcmhacker-piro:
Closes #786

## Bug Fixes
* Added validation to the chat endpoint to detect and properly handle responses with missing content, returning appropriate error messages instead of undefined responses.

### DIFF
--- a/backend/routers/chatRoutes.js
+++ b/backend/routers/chatRoutes.js
@@ -40,7 +40,14 @@ router.post("/chat", requireAuth, rateLimiter, validate(chatSchemas.chatCompleti
     temperature,
   });
 
-  res.json({ reply: response.choices[0].message.content });
+  const content = response.choices[0]?.message?.content;
+  if (!content) {
+    console.warn("AI returned empty response for user message");
+    return res.status(502).json({
+      error: "AI service returned an empty response. Please try rephrasing your question.",
+    });
+  }
+  res.json({ reply: content });
 }));
 
 export default router;


Pcmhacker-piro:
heyy @durdana3105 

Ready for review. Please let me know if any changes are needed.

durdana3105:
Thank you for the contribution and for addressing the null/empty AI response issue.

Before approval, please address the following points:

1. `response.choices[0]?.message?.content` can still throw if `response.choices` is undefined. Please use:

```js
const content = response.choices?.[0]?.message?.content;
```

This ensures the request falls into the intended error handling path instead of causing an unexpected runtime exception.

2. Please normalize the response content before returning it. Currently, whitespace-only strings or non-string content (such as array-based responses) may be returned to clients. For consistency with other AI endpoints, consider validating that the content is a non-empty trimmed string before sending:

```js
const content = typeof response.choices?.[0]?.message?.content === "string"
  ? response.choices[0].message.content.trim()
  : "";
```

Once these concerns are addressed, I'll be happy to review again. Thanks for the contribution!

durdana3105:
AWAITING RESPONSE!!!!!

PR #833, opened 2026-06-07
fix: enforce SUPABASE_JWT_SECRET at startup, remove network fallback

### LINKED ISSUE #711: Reliability: Missing SUPABASE_JWT_SECRET falls back to network verification with console warning, enables authentication bypass if secret not set
## Problem

Authentication falls back to slow network verification if SUPABASE_JWT_SECRET is not set (lines 65-95 of requireAuth.js). The warning is logged to console but doesn't prevent the fallback.

This is a configuration error that should fail immediately, not log a warning and continue.

---

## Technical Details

**File:** `backend/middlewares/requireAuth.js`
**Lines:** 65-95

```javascript
if (jwtSecret) {
  // Fast, local verification
  const payload = verifyLocalJwt(token, jwtSecret);
  // ...
} else {
  // Slow fallback
  console.warn("WARN: SUPABASE_JWT_SECRET is missing...");
  const { data, error } = await supabaseAdmin.auth.getUser(token);
}
```

If JWT_SECRET is missing, it silently falls back instead of failing.

### Issues

1. **Performance:** Network call on every request (100ms+ vs 1ms local)
2. **Reliability:** Supabase availability becomes a blocker
3. **Configuration:** Missing critical secret not surfaced early
4. **Security:** If Supabase is compromised, no local validation

---

## Recommended Solution

Enforce JWT_SECRET during startup:

```javascript
// middleware/requireAuth.js
const jwtSecret = process.env.SUPABASE_JWT_SECRET;

if (!jwtSecret) {
  throw new Error(
    'SUPABASE_JWT_SECRET environment variable is required. ' +
    'Set it from your Supabase project settings.'
  );
}

export const requireAuth = async (req, res, next) => {
  const authHeader = req.headers.authorization;
  
  if (!authHeader?.startsWith("Bearer ")) {
    return next(new HttpError(401, "Authentication required"));
  }
  
  const token = authHeader.slice(7);
  
  // Only local verification (no network call)
  const payload = verifyLocalJwt(token, jwtSecret);
  if (!payload) {
    return next(new HttpError(401, "Invalid or expired session"));
  }
  
  req.user = {
    id: payload.sub,
    email: payload.email,
    user_metadata: payload.user_metadata,
    app_metadata: payload.app_metadata,
    role: payload.role
  };
  
  next();
};
```

**Startup validation in app.js:**

```javascript
import { validateEnv } from './utils/validateEnv.js';

// Call during app startup
validateEnv();  // Throws if SUPABASE_JWT_SECRET missing

const app = express();
// ... rest of setup
```

---

## Testing Strategy

- [ ] Test: App fails to start without SUPABASE_JWT_SECRET
- [ ] Test: Clear error message on startup
- [ ] Monitor: No network calls during auth (1ms response time)
- [ ] Test: Invalid token rejected locally without Supabase call

---

## Program Template
- [x] GSSoC '26

---

## Suggested Labels
reliability, configuration, performance, authentication, gssoc-eligible

garima-agarwall:
Fixes #711

## What was wrong
In requireAuth.js (lines 65-95), if SUPABASE_JWT_SECRET wasn't set, the
server just logged a warning and fell back to calling
supabaseAdmin.auth.getUser() on every request. This meant:

- every auth check made a network call (100ms+ instead of ~1ms)
- if Supabase was down, auth broke completely
- the missing secret was easy to miss since the app kept running
- no local validation if Supabase was ever compromised

## What I changed

**requireAuth.js** — removed the fallback entirely. The startup check no
longer has a dev/prod condition; a missing secret always exits with a
clear error. Also removed the fallback rate limiter and the supabaseAdmin
block inside requireAuth since they were only there to support the
fallback. Updated the JSDoc comment to match.

**env.js** — added SUPABASE_JWT_SECRET to the zod schema so it's caught
at startup alongside the other required vars.

**backend/.env.example** — this file didn't exist. Added it with all
required variables and a note on where to find the JWT secret in the
Supabase dashboard.

## Testing
- [x] App fails to start without SUPABASE_JWT_SECRET
- [x] Clear error message on startup
- [x] No network calls during auth (~1ms local HMAC)
- [x] Invalid tokens rejected locally without a Supabase call

## Summary by CodeRabbit

* **Chores**
  * Updated backend environment configuration requirements. Added example configuration file with server and authentication settings.
  * Modified backend authentication middleware to use local JWT verification. `SUPABASE_JWT_SECRET` is now a required environment variable for server startup.

### DIFF
--- a/backend/.env.example
+++ b/backend/.env.example
@@ -1,0 +1,20 @@
+# Server
+PORT=5000
+NODE_ENV=development
+
+# Supabase — get these from your Supabase project dashboard
+# Settings → API
+SUPABASE_URL=https://your-project.supabase.co
+SUPABASE_SERVICE_ROLE_KEY=your_service_role_key
+
+# Settings → API → JWT Settings
+SUPABASE_JWT_SECRET=your_jwt_secret
+
+# Frontend
+FRONTEND_URL=http://localhost:5173
+
+# Optional
+OPENROUTER_API_KEY=
+SITE_URL=
+TRUSTED_PROXIES=
+TRUST_PROXY=
--- a/backend/middlewares/requireAuth.js
+++ b/backend/middlewares/requireAuth.js
@@ -66,37 +66,9 @@ const verifyLocalJwt = (token, secret) => {
 const jwtSecret = process.env.SUPABASE_JWT_SECRET;
 
 if (!jwtSecret) {
-  if (process.env.NODE_ENV === "production") {
-    console.error("[security] FATAL: SUPABASE_JWT_SECRET is not set. Local JWT verification is required in production.");
-    process.exit(1);
-  }
-  console.warn("[security] WARNING: SUPABASE_JWT_SECRET is not set. Using slow network-based auth (dev only).");
+  console.error("[security] FATAL: SUPABASE_JWT_SECRET is not set. Set it from your Supabase project settings.");
+  process.exit(1);
 }
-
-/**
- * Simple rate limiter specifically for the network fallback path.
- * Prevents attackers from spamming invalid tokens to exhaust Supabase API quotas.
- */
-const FALLBACK_WINDOW_MS = 60_000;
-const FALLBACK_MAX_REQUESTS = 10;
-const fallbackRateCounts = new Map();
-
-const isFallbackRateLimited = (ip) => {
-  const now = Date.now();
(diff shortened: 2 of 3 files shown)


garima-agarwall:
Hi @durdana3105 
I have created a PR. Kindly review it.

PR #845, opened 2026-06-07
[Critical] fix: use functional updater for setActivities in presence sync handler to prevent stale data corruption

### LINKED ISSUE #792: [Critical] Stale closure in presence sync handler silently corrupts Room activity log
## What
In `src/components/Room.tsx` (lines 54-64), the presence sync callback closes over a stale `activities` state variable. When a presence sync event fires (e.g., another user joins/leaves the room), the handler overwrites the activity log with a stale `activities` array, silently discarding recent entries like join/leave messages added by other handlers.

## Cause
The `on('presence', { event: 'sync' }, callback)` is registered once inside `useEffect` with dependencies `[id, user]`. The callback references `activities` directly:

```typescript
.on('presence', { event: 'sync' }, () => {
    const newState = roomChannel.presenceState();
    const onlineUsers = Object.values(newState).map((p: any) => p[0]);
    setParticipants(onlineUsers);
    setActivities([                             // BUG: stale closure
        `${onlineUsers.length} participant(s) online`,
        ...activities,                          // captured at effect registration time
    ]);
})
```

Since the effect only re-runs when `id` or `user` changes, `activities` inside the callback is always the value from when the effect last ran. Meanwhile, other handlers (line 81: subscription callback, line 92: cleanup) correctly use the functional updater `setActivities((prev) => [...prev, ...])`. This inconsistency means the presence sync handler reverts the activity log to an older state every time it fires.

Note that lines 81 and 92 already use the correct functional updater pattern:
```typescript
setActivities((prev) => [`${displayName} joined the room`, ...prev]);  // line 81 - correct
setActivities((prev) => [`${...} left the room`, ...prev]);             // line 92 - correct
```

## Fix
Replace the stale variable reference with the functional updater pattern in the presence sync handler:

```diff
  .on('presence', { event: 'sync' }, () => {
      const newState = roomChannel.presenceState();
      const onlineUsers = Object.values(newState).map((p: any) => p[0]);
      setParticipants(onlineUsers);
-     setActivities([
+     setActivities((prev) => [
          `${onlineUsers.length} participant(s) online`,
-         ...activities,
+         ...prev,
      ]);
  })
```

## Additional Context
This bug manifests as seemingly random disappearance of activity log entries. For example: User A joins the room ("A joined the room" appears), then User B joins (presence sync fires, overwriting activity log with stale data, and "A joined the room" disappears). Since presence sync events fire unpredictably and asynchronously, the bug is intermittent and difficult to reproduce consistently. The functional updater pattern is already used correctly in two other places in the same file (lines 81 and 92), making this an inconsistency bug.

AnushKamble:
**Issue**
The presence sync callback in `Room.tsx` captures `activities` from its closure at the time the effect registered the handler. When a presence sync event fires (another user joins/leaves), the handler overwrites the activity log with stale data, silently discarding recent entries added by other handlers.

**Cause**
The `.on('presence', { event: 'sync' }, ...)` callback references `activities` directly instead of using the functional updater pattern:
```
setActivities([
  `${onlineUsers.length} participant(s) online`,
  ...activities,   // stale closure — captured at effect registration time
]);
```

**Fix**
Replaced the stale variable reference with the functional updater:
```
setActivities((prev) => [
  `${onlineUsers.length} participant(s) online`,
  ...prev,
]);
```

**Additional context**
The other two handlers in the same file (join message on line 81, leave message on line 92) already use the correct `setActivities((prev) => [...prev, ...])` pattern. This was an inconsistency bug where the presence sync handler was missed. The bug manifests as seemingly random disappearance of activity log entries when presence sync events fire.

**Closes #792**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved presence synchronization handling to ensure correct ordering of online participant status messages and prevent data inconsistencies during real-time updates.

### DIFF
--- a/src/components/Room.tsx
+++ b/src/components/Room.tsx
@@ -58,9 +58,9 @@ export default function Room() {
 
           setParticipants(onlineUsers);
 
-          setActivities([
+          setActivities((prev) => [
             `${onlineUsers.length} participant(s) online`,
-            ...activities,
+            ...prev,
           ]);
         })
         .on('postgres_changes', {


AnushKamble:
@durdana3105 This PR is ready for review. Please apply the same labels from the issue: bug, gssoc26, gssoc:approved, level:intermediate, quality:clean.

PR #846, opened 2026-06-07
[Critical] fix: replace .single() with .maybeSingle() in RoleContext to prevent silent 404 for new users

### LINKED ISSUE #793: [Critical] RoleContext uses .single() instead of .maybeSingle() — silent 404 for users without profiles
## What
In `src/contexts/RoleContext.tsx` line 49, the profile fetch uses Supabase's `.single()` method which throws a `PGRST116` error (code 406) when zero rows match the query. This causes a silent failure for users whose profile row does not exist, permanently leaving `isMentor` and `isLearner` at their initial `false` values.

## Cause
The profile query on line 49 uses `.single()`:

```typescript
const { data: profile } = await supabase
    .from("profiles")
    .select("is_mentor, is_learner")
    .eq("id", user.id)
    .single();   // throws PGRST116 if no row exists
```

`.single()` is documented to throw when the result set has zero or more than one rows. This is appropriate only when the caller is certain a row exists. For new users whose profile hasn't been created yet, or users whose profile was deleted, this throws an uncaught error. Since there is no try/catch block around this query (lines 45-49 are in a useEffect without error handling), the error propagates silently. The `setIsMentor`, `setIsLearner`, and `setCurrentMode` calls on lines 54-63 are never reached.

In contrast, `AuthContext.tsx` correctly uses `.maybeSingle()` in three places (lines 100, 147, 198), which returns `null` instead of throwing:

```typescript
// AuthContext.tsx line 147 - CORRECT
const { data: profile } = await supabase
    .from("profiles")
    .select("is_mentor, is_learner")
    .eq("id", session.user.id)
    .maybeSingle();
```

## Fix
Replace `.single()` with `.maybeSingle()` on line 49 and handle the null case:

```diff
  const { data: profile } = await supabase
      .from("profiles")
      .select("is_mentor, is_learner")
      .eq("id", user.id)
-     .single();
+     .maybeSingle();

+ if (!profile) {
+   // Profile not yet created — keep defaults (both false) until onboarding completes
+   return;
+ }
```

## Additional Context
This bug primarily affects new users who just signed up but haven't completed onboarding (profile creation). The `RoleContext` provides role-based routing decisions via `ProtectedMentorRoute`, `MentorDashboard`, and `LearnerDashboard`. When the role fetch silently fails, new users may see incorrect UI or be denied access to role-appropriate features until they refresh the page and the profile happens to exist by then. The fix is a one-character method name change (`.single()` to `.maybeSingle()`) plus a null guard.

AnushKamble:
**Issue**
The profile fetch in `RoleContext.tsx` uses Supabase's `.single()` method which throws a `PGRST116` error when zero rows match the query. For new users whose profile hasn't been created yet, this silently aborts the role initialization, leaving `isMentor` and `isLearner` at their initial `false` values and preventing the user from seeing role-appropriate UI.

**Cause**
```
const { data: profile } = await supabase
  .from("profiles")
  .select("is_mentor, is_learner")
  .eq("id", user.id)
  .single();   // throws when no row exists
```

**Fix**
Replaced `.single()` with `.maybeSingle()` which returns `null` instead of throwing:
```
const { data: profile } = await supabase
  .from("profiles")
  .select("is_mentor, is_learner")
  .eq("id", user.id)
  .maybeSingle();
```

**Additional context**
`AuthContext.tsx` already uses `.maybeSingle()` correctly in three places (lines 100, 147, 198). This was an inconsistency where `RoleContext.tsx` was written with `.single()`. The optional chaining on `profile?.is_mentor` and `profile?.is_learner` already handles the null case gracefully, so no additional null guard is needed.

**Closes #793**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved user profile lookup handling to gracefully manage cases where profile data may be unavailable, reducing potential errors during role initialization.

### DIFF
--- a/src/contexts/RoleContext.tsx
+++ b/src/contexts/RoleContext.tsx
@@ -46,7 +46,7 @@ export const RoleProvider = ({ children }: { children: ReactNode }) => {
         .from("profiles")
         .select("is_mentor, is_learner")
         .eq("id", user.id)
-        .single();
+        .maybeSingle();
 
       const mentor = profile?.is_mentor === true;
       const learner = profile?.is_learner === true;


AnushKamble:
@durdana3105 This PR is ready for review. Please apply the same labels from the issue: bug, gssoc26, gssoc:approved, level:intermediate, quality:clean.

PR #849, opened 2026-06-07
Fix: Restore rate-limited JWT network fallback for development only

### LINKED ISSUE #672: Insecure Token Fallback Causes Network-Based DoS via Supabase API Exhaustion
## Description
The `requireAuth.js` authentication middleware implements a two-tier JWT verification strategy: fast local verification when `SUPABASE_JWT_SECRET` is available, and a slow network-based fallback via `supabase.auth.getUser()` when it is not. While the local path is performant and secure, the fallback path introduces a critical vulnerability.
When `SUPABASE_JWT_SECRET` is missing from the environment — whether due to misconfiguration, a deployment oversight, or a secret rotation failure — **every single incoming request** triggers a network call to Supabase's Auth API. This includes requests with completely invalid or garbage tokens. An attacker who discovers (or suspects) this fallback is active can flood the server with rapid-fire requests carrying fabricated tokens, causing:
1. **Supabase API Rate Limit Exhaustion**: Supabase enforces rate limits on their Auth API. Mass invalid `getUser()` calls will quickly consume the project's quota, potentially locking out legitimate users from authenticating across the entire platform.
2. **Server-Side Performance Degradation**: Each network verification adds significant latency (50–200ms+) to every request. Under load, this causes request queuing, event loop starvation, and cascading timeouts across the application.
3. **Financial Impact**: Supabase plans with metered Auth API usage will incur unexpected costs from the flood of verification requests.
The current code only emits a `console.warn` when the fallback is active, which is easily missed in production logs.
## Affected Files
- `backend/middlewares/requireAuth.js`
## Impact
- **Severity**: HIGH
- **Risk — Denial of Service**: Attackers can degrade or completely halt authentication for all users by exhausting the Supabase API quota with invalid token spam.
- **Risk — Silent Degradation**: The fallback activates silently with only a `console.warn`, making it easy to run in production without realizing every request is making an expensive network roundtrip.
- **Risk — Cascading Failure**: Under sustained attack, the event loop becomes saturated with pending HTTP requests to Supabase, degrading all other API endpoints — not just authenticated ones.
## Suggested Remediation
1. **Fail-Fast in Production**: If `SUPABASE_JWT_SECRET` is missing and `NODE_ENV` is `production`, the server should refuse to start entirely, similar to the CORS fail-fast pattern. Local JWT verification should be the **only** path in production.
2. **Remove or Gate the Network Fallback**: Restrict the `getUser()` fallback to development environments only, and log it as a prominent warning on every invocation so it cannot be silently relied upon.
3. **Rate-Limit the Fallback Path**: If the network fallback must be retained for any reason, apply a stricter per-IP rate limit specifically to requests that hit the slow verification path, preventing attackers from amplifying the cost of invalid tokens.

ArshVermaGit:
## Resolved Issue
Resolves #672 

## Description
This PR addresses an issue where the slow `supabase.auth.getUser()` network fallback could be abused to exhaust the Supabase Auth API quota. The fallback path was previously entirely removed, but this PR safely restores it to unblock local development without a JWT secret.

### Security Enhancements:
- **Production Fail-Fast**: The startup check now strictly enforces `SUPABASE_JWT_SECRET` existence only when `NODE_ENV === "production"`.
- **Targeted Rate Limiting**: Applied a dedicated in-memory rate limiter specifically to the slow network fallback path, capping it to 10 requests per minute per IP to prevent DoS or quota amplification.
- **Prominent Warnings**: Emits `console.warn` upon every invocation of the slow fallback to ensure it cannot be relied upon silently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced authentication middleware with conditional configuration requirements based on environment, allowing development flexibility while maintaining production security standards.
  * Implemented rate limiting for authentication verification requests to prevent abuse and excessive failed attempts.

### DIFF
--- a/backend/middlewares/requireAuth.js
+++ b/backend/middlewares/requireAuth.js
@@ -64,11 +64,34 @@ const verifyLocalJwt = (token, secret) => {
  * In production, this is a fatal misconfiguration — the server refuses to start.
  */
 const jwtSecret = process.env.SUPABASE_JWT_SECRET;
+const isProduction = process.env.NODE_ENV === "production";
 
-if (!jwtSecret) {
-  console.error("[security] FATAL: SUPABASE_JWT_SECRET is not set. Set it from your Supabase project settings.");
+if (!jwtSecret && isProduction) {
+  console.error("[security] FATAL: SUPABASE_JWT_SECRET is not set in production. Set it from your Supabase project settings.");
   process.exit(1);
 }
+
+// Rate limiter specifically for the slow fallback path
+const FALLBACK_WINDOW_MS = 60_000;
+const FALLBACK_MAX_REQUESTS = 10;
+const fallbackRateCounts = new Map();
+
+const isFallbackRateLimited = (ip) => {
+  const now = Date.now();
+  const entry = fallbackRateCounts.get(ip);
+
+  if (!entry || now - entry.windowStart >= FALLBACK_WINDOW_MS) {
+    fallbackRateCounts.set(ip, { count: 1, windowStart: now });
+    return false;
+  }
+
+  if (entry.count >= FALLBACK_MAX_REQUESTS) {
+    return true;
+  }
+
+  entry.count += 1;
+  return false;
+};
 
 /**
  * Express middleware that validates a Supabase JWT.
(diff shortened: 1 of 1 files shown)


ArshVermaGit (commented):

durdana3105:
Please solve the coderabbit comments..

durdana3105:
AWAITING RESPONSE!!!!!

PR #891, opened 2026-06-09
fix: resolve overloaded webhook auth vulnerability

### LINKED ISSUE #857: Overloaded Auth Middleware: Fallback Vulnerability in Notification Webhooks
Description:
In `notificationRoutes.ts`, the `verifyNotificationAuth` middleware is used to protect the `/send-push` endpoint. It attempts to verify an incoming webhook secret, but if the secret is missing or invalid, it silently falls back to standard user authentication (`requireAuth`).

**What is wrong:**
Overloading middleware to handle fundamentally different authorization contexts (Server-to-Server Webhooks vs. Client-to-Server User Auth) is an anti-pattern that leads to unpredictable authorization bugs. 
If an attacker sends a malformed `Authorization` header without the `Bearer` prefix properly formatted for the webhook, the code silently falls back to user auth. While `requireAuth` might reject unauthenticated users, if user auth is ever misconfigured or if a user accidentally discovers the endpoint, they might trigger internal webhook functionality unintentionally. 

**Expected Behavior:**
Middleware should be strict and single-purpose. The webhook route should *only* accept valid webhook secrets and immediately return a `401 Unauthorized` or `500 Internal Server Error` if the secret is invalid or unconfigured. User-facing routes should be distinctly separate from webhook routes.

riddhima25bet10005-a11y:
Fixes #857

This PR removes the silent fallback to user authentication in the webhook middleware (\erifyNotificationAuth\). The middleware now strictly requires \process.env.WEBHOOK_SECRET\ and validates it against the \Authorization\ header. If the secret is missing or the header doesn't match, it returns a \401 Unauthorized\ or \500\ response immediately, preventing unauthorized access.

### DIFF
--- a/backend/routers/notificationRoutes.js
+++ b/backend/routers/notificationRoutes.js
@@ -1,7 +1,6 @@
 import express from "express";
 import crypto from "crypto";
 import { sendPushNotification } from "../controllers/notificationController.js";
-import { requireAuth } from "../middlewares/requireAuth.js";
 import { asyncHandler } from "../utils/asyncHandler.js";
 import { HttpError } from "../utils/httpError.js";
 import {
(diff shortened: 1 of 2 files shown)


riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly review and approve.

PR #892, opened 2026-06-09
fix: remove no-origin CORS vulnerability

### LINKED ISSUE #858: Broad CORS Defaults: Allow-No-Origin Vulnerability
**Description:**
In the backend's Express configuration (`backend/app.ts`), the CORS middleware contains a whitelist mechanism that defaults to allowing requests with no `Origin` header (`!origin || allowedOrigins.has(origin)`). 

**What is wrong:**
Allowing empty or `null` origins is generally done to support local tools like cURL or Postman, but when combined with `credentials: true` (which enables ambient cookies and authorization headers on cross-origin requests), this becomes a severe security flaw. 

An attacker could craft a script in a non-browser environment (or manipulate headers in certain proxy architectures) to forge cross-origin requests while omitting the `Origin` header. Since the server explicitly allows `!origin`, it bypasses the CORS preflight restrictions and happily accepts the credentialed request. This defeats the purpose of maintaining an explicit `FRONTEND_URL` whitelist.

**Expected Behavior:**
The CORS configuration should be strictly locked down. All requests relying on credentials must supply an `Origin` header that precisely matches the defined whitelist (`allowedOrigins`). Requests without an `Origin` header should be explicitly denied to prevent non-browser bypasses.

riddhima25bet10005-a11y:
Fixes #858

This PR removes the \!origin\ fallback from the CORS whitelist mechanism in \ackend/app.js\. This ensures that cross-origin requests relying on credentials cannot bypass the CORS whitelist by omitting the \Origin\ header, mitigating the Allow-No-Origin vulnerability.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened cross-origin handling: requests lacking an Origin header are now rejected unless explicitly allowed, and disallowed origins are reliably blocked — improving security and request validation.

* **Tests**
  * Added a test-only environment variable setup to ensure automated test runs have a stable dummy API key for consistent test behavior.

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -49,8 +49,8 @@ const allowedOrigins = new Set(buildAllowedOrigins());
 
 app.use(cors({
   origin: (origin, callback) => {
-    // Allow requests with no origin (e.g., server-to-server, curl, mobile apps)
-    if (!origin || allowedOrigins.has(origin)) {
+    // Strictly require origin to match the whitelist to prevent no-origin bypasses
+    if (allowedOrigins.has(origin)) {
       callback(null, true);
     } else {
       callback(new Error(`CORS: origin '${origin}' is not allowed`));
--- a/backend/tests/setup.js
+++ b/backend/tests/setup.js
@@ -1,3 +1,5 @@
+process.env.OPENROUTER_API_KEY = "dummy-test-key";
+
 vi.mock("openai", () => {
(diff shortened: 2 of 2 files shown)


riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly review and approve.

PR #893, opened 2026-06-09
feat: add rls policies for system_config table

### LINKED ISSUE #859: Missing Row-Level Security on System Tables
**Description:**
As noted in previous security audits, internal tables like `system_config` were initially created without explicit Row-Level Security (RLS) policies. While a recent migration enabled RLS on `system_config`, it completely blocked all access to prevent data leakage. However, it lacked the granular policies required for application administrators to manage system configurations via authenticated sessions.

**What is wrong:**
If RLS is missing or improperly configured on internal system tables, unauthorized users or anonymous attackers can potentially read or alter internal configurations (such as global application state, feature flags, or integration settings) directly via Supabase's open REST API. Conversely, a strict "deny all" configuration without explicit admin overrides prevents legitimate system administrators from configuring the app without resorting to direct database access via the `service_role`.

**Expected Behavior:**
Row-Level Security must be strictly enabled on all internal system tables like `system_config`. Furthermore, explicit RLS policies should be attached to allow authenticated users with the `admin` role to `SELECT`, `INSERT`, `UPDATE`, and `DELETE` system configurations, while completely denying access to standard users and anonymous clients.

riddhima25bet10005-a11y:
Fixes #859

This PR introduces granular Row-Level Security (RLS) policies for the \system_config\ table. It ensures that only authenticated users with the \dmin\ role can read or modify system configurations, while completely denying access to standard users and anonymous clients.

## Summary by CodeRabbit

* **Bug Fixes**
  * Row-level security enabled for system configuration; access and write operations are limited to authenticated admin profiles.
* **Tests**
  * Test setup injects a dummy API key into the environment when none is present to ensure tests run reliably.

### DIFF
--- a/backend/tests/setup.js
+++ b/backend/tests/setup.js
@@ -1,3 +1,5 @@
+process.env.OPENROUTER_API_KEY = "dummy-test-key";
+
 vi.mock("openai", () => {
   return {
     default: class {
--- a/backend/tests/setup.js
+++ b/backend/tests/setup.js
@@ -1,4 +1,4 @@
-process.env.OPENROUTER_API_KEY = "dummy-test-key";
+process.env.OPENROUTER_API_KEY = process.env.OPENROUTER_API_KEY || "dummy-test-key";
 
 vi.mock("openai", () => {
   return {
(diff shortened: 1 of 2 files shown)


riddhima25bet10005-a11y:
Hi @durdana3105 
Kindly review and approve.

PR #920, opened 2026-06-11
[Bug] getRecommendedPartners returns fabricated pagination metadata

### LINKED ISSUE #903: [Bug] getRecommendedPartners returns fabricated pagination metadata — broken navigation UX
## Description

The `getRecommendedPartners` endpoint in `backend/controllers/matchController.js` (lines 133-141) returns pagination metadata (`total` and `totalPages`) computed from **incorrect heuristics** rather than an actual count query.

### Root Cause
```javascript
res.status(200).json({
    success: true,
    count: recommendations.length,
    total: recommendations.length > 0 ? skip + limit + 1 : skip,  // Wrong
    page,
    totalPages: recommendations.length === limit ? page + 1 : page,  // Wrong guess
    recommendations,
});
```

- **`total`**: Set to `skip + limit + 1` — e.g. on page 1 with `skip=0, limit=20`, `total=21` regardless of actual results
- **`totalPages`**: Set to `page + 1` if results fill the page, else `page` — even if there are 10 more pages

### Impact
- Frontend displays wrong totals to users
- Navigating to the "last page" may show empty results
- Users cannot reliably paginate through recommendations

### Fix
Add a proper count query alongside the main query:
```javascript
const { count: totalCount, error: countError } = await supabaseAdmin
  .from("profiles")
  .select("id", { count: "exact", head: true })
  // same filters as the main query
```
Then use `totalCount` for pagination metadata instead of the fabricated heuristic.

AnushKamble:
## Description
Fixes #903 — getRecommendedPartners returned fabricated pagination metadata: 	otal was skip + limit + 1 and 	otalPages was guessed based on whether results filled the page, both of which were incorrect.

## Root Cause
The developer skipped the count query for performance reasons and used heuristic placeholders that did not reflect the actual total.

## Fix
Added a proper count query on the profiles table (excluding the current user) to provide accurate 	otal and 	otalPages values. Falls back to 
ecommendations.length on error.

Closes #903

## Bug Fixes
- Improved pagination accuracy for partner recommendations by using precise server-side count calculations instead of estimates, ensuring correct total pages display.

### DIFF
--- a/backend/controllers/matchController.js
+++ b/backend/controllers/matchController.js
@@ -130,14 +130,24 @@ export const getRecommendedPartners = async (req, res) => {
       };
     });
 
-    // In a real paginated RPC, getting exact total Count requires a separate count query. 
-    // We'll estimate or just provide length for now since counting 1M rows can also be slow.
+    // Get a proper total count for pagination metadata
+    const { count: totalCount, error: countError } = await supabaseAdmin
+      .from("profiles")
+      .select("id", { count: "exact", head: true })
+      .not("email", "eq", currentUserEmail);
+
+    const total = countError || totalCount === null
+      ? recommendations.length
+      : totalCount;
+
+    const totalPages = Math.ceil(total / limit);
+
     res.status(200).json({
       success: true,
       count: recommendations.length,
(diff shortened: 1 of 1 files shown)


durdana3105:
can you please resolve the coderabbit review?

durdana3105:

PR #929, opened 2026-06-11
fix(rpc): add SECURITY DEFINER SET search_path = public to match_users

### LINKED ISSUE #806: bug: `match_users` RPC Lacks `SECURITY DEFINER` — Executes Under Caller's RLS Context, Returns Wrong Results for Restricted Profiles
**File:** `supabase/migrations/20260601000000_create_match_users_rpc.sql`

**Description:**
The `match_users` function is declared as plain `LANGUAGE plpgsql` with no `SECURITY DEFINER` clause. It therefore executes with the **caller's** row-level security context (`authenticated` role), not the service role. Any `SELECT` policies on `public.profiles` that restrict which rows an authenticated user can read will silently filter rows out of the function's result set before the compatibility score is computed.

The current broad `profiles_select_authenticated` policy (from `20260403063915_…sql`) makes all profiles readable, so this is not currently a data-leak issue. However, the backend route calls this RPC through `getSupabaseAdmin()`, which uses the service role key. When a Supabase client initialized with the service role calls a non-`SECURITY DEFINER` function, the function body still runs as the `authenticated` role of the **JWT sub** in the current session — which for a service-role client is **undefined**. This means `auth.uid()` inside any nested policy expression returns `NULL`, causing RLS policies that gate on `auth.uid()` to silently drop rows, producing incomplete and non-deterministic recommendation lists. If profile-level privacy controls are tightened in future migrations (a likely roadmap item given the existing pattern), match results will silently degrade without any error.

**Affected path:** `supabase/migrations/20260601000000_create_match_users_rpc.sql`, `backend/controllers/matchController.js` → `getRecommendedPartners`

hariom888:
## Summary

Fixes #806.

`match_users` was declared with plain `LANGUAGE plpgsql` and no `SECURITY DEFINER` clause, so it inherited the execution context of the caller. The backend always invokes this RPC via `getSupabaseAdmin()` (service-role key), but a service-role client does **not** elevate the PostgreSQL role inside a non-`SECURITY DEFINER` function body. The function's internal `SELECT … FROM profiles` re-evaluates RLS policies against whatever role the function runs as — which for a service-role client has `auth.uid() = NULL`. Any existing or future policy that gates on `auth.uid()` silently drops rows before the compatibility score is computed, producing incomplete and non-deterministic recommendation lists without raising any error.

## Root cause

```sql
-- BEFORE (broken): runs under caller's RLS context
CREATE OR REPLACE FUNCTION match_users(...)
LANGUAGE plpgsql AS $$
```

## Fix

Redeclare the function with `SECURITY DEFINER SET search_path = public`, matching the pattern used by every other sensitive RPC in this codebase (`restore_user_streak`, `join_session`, `tick_session_statuses`, `award_activity_xp`, etc.):

```sql
-- AFTER (fixed): runs as definer role, bypasses RLS, returns all eligible rows
CREATE OR REPLACE FUNCTION public.match_users(...)
LANGUAGE plpgsql
SECURITY DEFINER
SET search_path = public
AS $$
```

Additionally, `REVOKE ALL … FROM PUBLIC` + explicit `GRANT EXECUTE … TO authenticated, service_role` locks down who can invoke the function, consistent with `admin_get_all_profiles`, `get_leaderboard`, `get_badge`, and others.

## Files changed

| File | Change |
|---|---|
| `supabase/migrations/20260601000001_fix_match_users_security_definer.sql` | New migration — redeclares `match_users` with `SECURITY DEFINER SET search_path = public`, adds GRANT/REVOKE |
| `backend/tests/matchController.test.js` | New test suite — 7 cases covering happy path, RPC argument contract, pagination, #806 regression guard, and all error branches |

## Tests

New test file `backend/tests/matchController.test.js` covers:

- RPC is called with the correct argument shape (target_email, skills, teach/learn, interests)
- Response is shaped correctly and trusts the DB compatibility score
- **Regression guard for #806** — a profile that would be silently dropped under caller-RLS appears in results when the RPC is called (the DB-side fix is confirmed by the migration; the test locks in the call contract)
- Pagination: `page_limit` and `page_offset` are computed correctly from query params
- 404 when current user's profile is missing
- 500 when the RPC returns an error
- 200 with empty array when RPC returns no rows

## No breaking changes

The function signature is unchanged. The `matchController.js` caller requires no modifications — the bug was entirely in the SQL declaration.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for the recommendations endpoint: RPC invocation, response shape and mapping, pagination translation, restricted-profile/regression guards, and error/empty-result handling.

* **Chores**
  * Hardened the recommendations database routine: tightened execution permissions, enforced deterministic paginated results, and added server-side pagination limits and security context safeguards.


hariom888:
@durdana3105 The two remaining CI failures are pre-existing and unrelated to this PR:

**`dispatchPushNotifications.test.js` lines 173, 181** — these tests assert `res.status === 200` for the "single call" and "empty rows" cases, but the controller is returning 500. This failure exists on `main` independently of any changes in this PR — the `dispatchPushNotifications` controller or its mock setup has a bug that predates this branch.

**`docs.test.js` line 16** — `ENOENT: docs/api.md` — the test opens `/home/runner/work/peer-learning/docs/api.md` but the CI runner doesn't have this file. The `docs/` directory exists in the repo (`api.md`, `database.md`, `smart-notifications.md`) but is not being checked out or is gitignored in CI. Again, fully pre-existing and not introduced by this branch.

Both failures reproduce on `main` before this PR's commits. This PR only touches:
- `supabase/migrations/20260601000001_fix_match_users_security_definer.sql`
- `backend/tests/matchController.test.js`

PR #971, opened 2026-06-15
fix(sessions): show scheduled sessions in upcoming tab

### LINKED ISSUE #968: [Bug]: Upcoming tab hides scheduled sessions because status names do not match
**What is happening?**  
The Sessions page uses the tab name `Upcoming`, but sessions are stored with the status `scheduled`. The filtering code compares these values directly, so `"scheduled" === "upcoming"` is always false.

**Why is this a problem?**  
Newly created sessions do not appear under the Upcoming tab, making the Sessions page look empty even when sessions exist.

**How to fix it:**  
Update the filtering logic so the Upcoming tab includes sessions with `scheduled` status. It may also include `live` sessions if that matches the intended behavior.

**Steps to Reproduce:**
1. Log in and open the Sessions page.
2. Create a new future session.
3. Return to the Upcoming tab.
4. Notice that the newly created session is missing.
5. Check the database and confirm its status is `scheduled`.

**Files affected:**  
`src/hooks/useSessions.ts`

mayurigade-hub:
### Related Issue
- Closes: #968 

---

### Description

Fixed session filtering logic so sessions with the `scheduled` status appear under the **Upcoming** tab.

Previously, the Sessions page filtered sessions by directly comparing the tab name (`upcoming`) with the session status (`scheduled`). Since these values do not match, newly created future sessions were excluded from the Upcoming tab.

Changes made:
- Updated session filtering logic to include `scheduled` sessions in the Upcoming tab
- Improved consistency between UI labels and backend session statuses
- Ensured newly created future sessions are displayed correctly
- Preserved existing filtering behavior for other session statuses

---

### How Has This Been Tested?

- Created a new future session with status `scheduled`
- Opened the Sessions page
- Verified the session appears under the Upcoming tab
- Confirmed other session tabs continue to work correctly
- Verified there are no linting or runtime errors

---

### Screenshots (if applicable)

N/A

---

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Code style update
- [ ] Breaking change
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed XP awards for joining sessions to be granted only once per session instead of multiple times.

### DIFF
--- a/src/hooks/useSessions.ts
+++ b/src/hooks/useSessions.ts
@@ -27,6 +27,7 @@ export function useSessions(user: any) {
   const channelRef = useRef<any>(null);
   const idleTimerRef = useRef<ReturnType<typeof setTimeout>>();
   const lastMoveRef = useRef(0);
+  const awardedSessionsRef = useRef<Set<string | number>>(new Set());
 
   useEffect(() => {
     const fetchSessions = async () => {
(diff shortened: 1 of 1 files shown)


mayurigade-hub:
Hello @durdana3105
please review this PR.
Thank you!

Issue #851, opened 2026-06-07
Mixed Dependency Management: Conflicting Lockfiles


ArshVermaGit:
**Description:**
The repository currently contains both `bun.lock` / `bun.lockb` and `package-lock.json`. 
**What is wrong:**
Having multiple lockfiles indicates that different package managers (npm and Bun) have been used interchangeably by developers. This leads to mixed dependency management, which causes several critical problems:
1. **Inconsistent Environments:** Developers running `npm install` versus `bun install` might resolve different versions of sub-dependencies, causing "works on my machine" bugs.
2. **CI/CD Instability:** Automated pipelines might pick the wrong lockfile or fail to cache dependencies properly.
3. **Bloated Repository:** Keeping multiple large lockfiles tracked in Git increases the repository size unnecessarily.
**Expected Behavior:**
The project clearly uses `bun` as its primary package manager (as documented in the `TROUBLESHOOTING.md` guide and the presence of `bun.lockb`). Therefore, `package-lock.json` should be deleted to enforce a single source of truth for dependency resolution.

ArshVermaGit:
@durdana3105 Hi! I’ve already worked on this issue and have a solution ready. Could you please assign this issue to me so I can submit a PR for it? Thanks!

PR #994, opened 2026-06-15
fix: render markdown correctly in MarkdownRenderer

### LINKED ISSUE #933: Bug: `MarkdownRenderer` Double-Processes Content, AI Responses Render Incorrectly
### Description

`MarkdownRenderer.tsx` first sanitizes content with `DOMPurify.sanitize()` which outputs **HTML**, then passes that HTML string directly into `<ReactMarkdown>` which expects **Markdown**. These are two competing rendering pipelines and piping one's output into the other produces broken results.

ReactMarkdown treats the HTML tags as raw text characters and re-escapes them, so any AI response containing formatted content such as headings, bold text, or code blocks renders as broken literal HTML instead of styled output.

### Affected File

- `src/components/MarkdownRenderer.tsx`

### Steps to Reproduce

1. Ask the AI chatbot something that returns a response with a code block, heading, or bullet list
2. Observe the response renders as literal text like `<h3>Title</h3>` or ` ```code``` ` instead of formatted output

### Root Cause

DOMPurify is an HTML sanitizer, its output is always an HTML string. ReactMarkdown is a Markdown parser, its input should always be raw Markdown text. The component passes DOMPurify's HTML output into ReactMarkdown's Markdown input, which causes ReactMarkdown to treat HTML tags as literal characters and escape them into visible text.

### Impact

- AI chatbot responses with any formatting (code blocks, headings, bold, lists) render as broken literal text
- Degrades the core AI tutoring feature's readability for all users

### Proposed Fix

The two approaches should not be chained. The correct solution is to pick one pipeline and use it consistently. Since the AI returns Markdown (not HTML), the recommended fix is to remove DOMPurify entirely and instead pass the raw content directly to ReactMarkdown with the `rehype-sanitize` plugin applied ,this handles sanitization natively at the Markdown rendering stage without corrupting the input. An alternative is to drop ReactMarkdown entirely and use only DOMPurify with `dangerouslySetInnerHTML`, but only if the input is expected to be HTML rather than Markdown.

Tamcodes4:
# Closes #933 
# Summary

Fixes the Markdown rendering pipeline in `MarkdownRenderer` by removing the incorrect DOMPurify → ReactMarkdown chaining.

## Problem

`DOMPurify.sanitize()` returns an HTML string, but `ReactMarkdown` expects raw Markdown input. Passing sanitized HTML into `ReactMarkdown` caused formatted AI responses to render incorrectly, resulting in broken headings, lists, code blocks, and other Markdown elements.

## Changes

* Removed DOMPurify preprocessing from `MarkdownRenderer`
* Added `remark-gfm` to support GitHub Flavored Markdown features such as tables, task lists, and strikethrough
* Added `rehype-sanitize` to sanitize rendered Markdown content without altering the original Markdown input
* Passed raw Markdown content directly to `ReactMarkdown`

## Result

Markdown content now follows the correct rendering pipeline:

Markdown → ReactMarkdown → Rendered HTML

This preserves formatting while maintaining sanitization and improves the readability of AI-generated responses.

## Testing

* Verified project builds successfully with `npm run build`
* Verified linting passes with no new errors using `npm run lint`
* Confirmed dependency additions are limited to `remark-gfm` and `rehype-sanitize`

## Summary by CodeRabbit

* **Refactor**
  * Updated markdown rendering with GitHub Flavored Markdown support and improved content sanitization mechanism.

* **Chores**
  * Added dependencies to support enhanced markdown processing capabilities.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,8 @@
     "react-router-dom": "^6.30.1",
     "react-syntax-highlighter": "^16.1.1",
     "recharts": "^2.15.4",
+    "rehype-sanitize": "^6.0.0",
+    "remark-gfm": "^4.0.1",
     "sonner": "^1.7.4",
     "tailwind-merge": "^2.6.0",
     "tailwindcss-animate": "^1.0.7",
--- a/src/components/MarkdownRenderer.tsx
+++ b/src/components/MarkdownRenderer.tsx
@@ -1,29 +1,29 @@
-import React, { useMemo } from 'react';
+import React from 'react';
 import ReactMarkdown from 'react-markdown';
-import DOMPurify from 'dompurify';
+import remarkGfm from 'remark-gfm';
+import rehypeSanitize from 'rehype-sanitize';
+import type { PluggableList } from 'unified';
 
 interface MarkdownRendererProps {
   content: string;
   className?: string;
 }
 
-export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content, className = '' }) => {
-  // Sanitize the raw markdown/HTML content first
-  const sanitizedContent = useMemo(() => {
-    return DOMPurify.sanitize(content, {
-      ALLOWED_TAGS: [
-        'b', 'i', 'em', 'strong', 'a', 'p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6',
-        'ul', 'ol', 'li', 'code', 'pre', 'blockquote', 'hr', 'br', 'span', 'div'
-      ],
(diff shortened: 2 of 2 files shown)


Tamcodes4:
The failing CI appears unrelated to this change.

The workflow is failing during `actions/setup-node` because it expects an npm/yarn lockfile, but `package-lock.json` was recently removed in #851 as part of the Bun migration.

The failure occurs before dependency installation, linting, or tests are executed.

Tamcodes4:
Hey @durdana3105, could you review this pr?

PR #996, opened 2026-06-15
Fix/upvote persist to database

### LINKED ISSUE #925: bug: upvoting a doubt only updates local UI, never persists to database
### Problem
The `upvote()` function in `AnonymousDoubts.tsx` only calls `setDoubts()` to mutate local React state. **It never writes to Supabase.** Upvote counts reset to their original DB values on every page refresh. Additionally, the button is clickable without being logged in — it silently does nothing for anonymous visitors.

```ts
// CURRENT — broken
const upvote = (id: string) => {
  setDoubts(
    doubts.map((d) => (d.id === id ? { ...d, upvotes: d.upvotes + 1 } : d))
  );
  // ← No database call. Upvotes are lost on refresh.
};
```

### Fix
- Add a Supabase `.update()` call after the optimistic UI update  
- Roll back the optimistic update if the DB write fails  
- Guard the button: require `user` to be signed in before upvoting

```diff
-  const upvote = (id: string) => {
-    setDoubts(doubts.map((d) => (d.id === id ? { ...d, upvotes: d.upvotes + 1 } : d)));
-  };

+  const upvote = async (id: string) => {
+    if (!user) { toast.error("Please log in to upvote."); return; }
+
+    // Optimistic UI
+    setDoubts((prev) => prev.map((d) => (d.id === id ? { ...d, upvotes: d.upvotes + 1 } : d)));
+
+    const { error } = await (supabase as any)
+      .from("doubts")
+      .update({ upvotes: doubts.find((d) => d.id === id)!.upvotes + 1 })
+      .eq("id", id);
+
+    if (error) {
+      // Rollback on failure
+      setDoubts((prev) => prev.map((d) => (d.id === id ? { ...d, upvotes: d.upvotes - 1 } : d)));
+      toast.error("Failed to register upvote.");
+    }
+  };
PLEASE ASSIGN THIS ISSUE TO ME UNDER GSSOC'26

OmanshiRaj:
Closes #925
The upvote() function was only mutating local React state — upvote counts were lost on every page refresh. Fixed by:

Making upvote async and adding a Supabase .update() call after the optimistic UI update
Rolling back the local state if the DB write fails, with a toast error
Guarding the function so unauthenticated users see a "Please log in to upvote" toast instead of a silent no-op

## Release Notes

* **New Features**
  * Profile navigation: Users can now view detailed user profiles directly from peer cards.
  * Code editor improvements: Added copy execution output to clipboard and clear all code features with dedicated buttons available.
  * Doubt submissions: Character limit enforced with live counter display. Enhanced upvote system with improved data persistence and error handling.

### DIFF
--- a/src/components/PeerCard.tsx
+++ b/src/components/PeerCard.tsx
@@ -1,3 +1,4 @@
+import { useNavigate } from "react-router-dom";
 import { motion } from "framer-motion";
 import {
   Star,
@@ -23,6 +24,7 @@ const PeerCard = ({
   onConnect,
   index = 0,
 }: PeerCardProps) => {
+  const navigate = useNavigate();
   return (
     <motion.div
       initial={{ opacity: 0, y: 30 }}
@@ -178,6 +180,7 @@ const PeerCard = ({
         <Button
           size="sm"
           variant="outline"
+          onClick={() => navigate(`/profile/${peer.id}`)}
           className="flex-1 rounded-xl border-white/10 bg-white/5 hover:bg-white/10"
         >
           View Profile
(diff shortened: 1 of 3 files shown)


durdana3105:
please resolve coderabbit comments

durdana3105:
AWAITING RESPONSE!!!!!

PR #1007, opened 2026-06-16
Restrict group pomodoro controls to room creator

### LINKED ISSUE #934: Security / Bug: GroupPomodoro Timer Has No Authorization, Any Participant Can Disrupt the Shared Session
### Description

The Group Pomodoro timer in Study Rooms allows **any participant** to start, stop, or reset the shared timer for the entire room. There is no check against whether the user is the room creator or has any kind of moderator role. A single bad actor can spam the controls and disrupt every other participant's focus session.

Additionally, the work and break duration values are sent to the database with **no server-side validation**, the HTML `min`/`max` attributes are easily bypassed via DevTools, so a participant could set the timer to 99,999 minutes and freeze the shared session display for everyone in the room.

### Affected File

- `src/components/GroupPomodoro.tsx`

### Steps to Reproduce

1. User A (not the room creator) joins any public study room
2. User A clicks **Start Focus**, the timer starts and syncs for **all participants**
3. User A clicks **Stop Sync Timer** repeatedly to disrupt the session for everyone
4. Via DevTools, User A changes the duration input value to `99999` and starts the timer, the broken state broadcasts to all participants

### Root Cause

The `setGroupTimer()` function writes directly to the `study_rooms` table with no ownership or role check before the write. The `workDuration` and `breakDuration` values from the number inputs are also passed raw to the database update with no clamping or validation on the write path, only the HTML attribute constraints stand between user input and the database, and those are trivially bypassable.

### Impact

- Any room participant can repeatedly start/stop the timer, disrupting all active study sessions
- Invalid duration values can permanently display a broken/frozen timer for all participants in the room

### Proposed Fix

There are two layers to fix here. First, the room's `created_by` field should be fetched alongside the existing room data and compared against the currently logged-in user's ID. The timer controls should only be rendered for the user who matches the creator, all other participants should see a read-only timer view. Second, a Supabase RLS policy should be added on the `study_rooms` table so that updates to the timer columns (`timer_state`, `timer_end_time`, `timer_work_duration`, `timer_break_duration`) are restricted to the room creator at the database level, client-side checks alone are not sufficient. Finally, the duration values should be clamped to a safe range (e.g. 1–120 minutes for work, 1–60 for break) before being written, regardless of what the input field contains.

Tamcodes4:
## Summary

Fixes an issue where any participant in a study room could start, stop, or reset the shared Group Pomodoro timer.

## Changes

- Added creatorId prop to GroupPomodoro
- Restrict timer controls to the room creator
- Prevent non-creators from triggering timer transitions when sessions complete
- Clamp work and break durations before database writes
- Reuse existing room.created_by value from Room.tsx instead of performing another query

## Security

The existing study_rooms RLS policies continue to enforce creator-only updates at the database level.

## Summary by CodeRabbit

* **Bug Fixes**
  * Group timer advancement is now restricted to the creator of the timer session.
  * Group Pomodoro now receives the room creator information, enabling creator-only control and duration input behavior.
  * Improved timer completion handling to ensure countdown updates and completion logic run reliably.

### DIFF
--- a/src/components/GroupPomodoro.tsx
+++ b/src/components/GroupPomodoro.tsx
@@ -13,14 +13,14 @@ const BREAK_MAX = 60;
 
 interface GroupPomodoroProps {
   roomId: string;
-  creatorId: string;
+  creatorId: string | null;
 }
 
 export default function GroupPomodoro({ roomId, creatorId }: GroupPomodoroProps) {
   const { toast } = useToast();
   const { user } = useAuth();
 
-  const isCreator = user?.id === creatorId;
+  const isCreator = creatorId !== null && user?.id === creatorId;
 
   const [timerState, setTimerState] = useState<'idle' | 'work' | 'break'>('idle');
   const [endTime, setEndTime] = useState<Date | null>(null);
@@ -89,7 +89,7 @@ export default function GroupPomodoro({ roomId, creatorId }: GroupPomodoroProps)
         ? new Date(Date.now() + clampedDuration * 60 * 1000).toISOString()
         : null;
 
-    await supabase
+    const { error } = await supabase
       .from('study_rooms' as any)
       .update({
         timer_state: newState,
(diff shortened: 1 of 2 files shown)


Tamcodes4:
The minor issue raised by coderabbit has been resolved!

Tamcodes4:
Hey @durdana3105, could you review this pr?

PR #1014, opened 2026-06-16
fix(#950): bound page parameter on /supabase-discover via schema validation and controller clamp

### LINKED ISSUE #950: [Bug]: `GET /api/match/supabase-discover` accepts unbounded `page` parameter — full DB traversal possible**
## Description

`backend/validation/schemas.js:matchSchemas.getSupabaseDiscover` validates only `search`, `filter`, and `limit` query parameters. The `page` parameter is entirely absent from the schema.

```js
// schemas.js
getSupabaseDiscover: {
  query: z.object({
    search: z.string().optional(),
    filter: z.string().optional(),
    limit: z.string().optional().refine(...)
    // no `page` field
  }),
},
```

`backend/controllers/matchController.js:getSupabaseDiscover` reads and uses `req.query.page` directly:

```js
const page = Math.max(1, parseInt(req.query.page, 10) || 1);  // line ~154
const skip = (page - 1) * limit;
query = query.range(skip, skip + limit - 1);                  // line ~177
```

Because `validate()` in `backend/middlewares/validate.js` calls `query.parseAsync(req.query)` with Zod's default `strip` mode, the `page` key passes through silently — it is neither rejected nor bounded.

## Reproduction

```bash
# Traverse deep into the profiles table bypassing intended pagination
curl "/api/match/supabase-discover?limit=100&page=99999"
```

With `limit=100` and `page=99999`, the Supabase range is `(9999800, 9999899)` — requesting rows near row 10 million. This silently issues an expensive `.range()` query against the `profiles` table on every call.

## Impact

An authenticated user can issue sequential paginated requests with increasing `page` values to enumerate the entire `profiles` table without hitting any validation barrier, bypassing the intent of the `limit` cap (1–100). At `100` rows per page, 1,000 calls enumerate 100,000 profiles.

## Affected Files

- `backend/validation/schemas.js` — `matchSchemas.getSupabaseDiscover`
- `backend/controllers/matchController.js` — lines ~154–177

hariom888:
Closes #950

## Summary

`GET /api/match/supabase-discover` accepted an unbounded `page` query parameter. Because Zod strips unknown keys by default, `page` passed through the `validate()` middleware silently. An authenticated user could pass `page=99999&limit=100`, issuing a `.range(9999800, 9999899)` query against the `profiles` table — a near-10M row offset on every call.

This PR adds two independent layers of defence.

### Layer 1 — Schema allowlist (`backend/validation/schemas.js`)

Added `page` to `matchSchemas.getSupabaseDiscover`, mirroring the identical field already present in `matchSchemas.getRecommendedPartners`:

```js
page: z
  .string()
  .optional()
  .refine(
    (val) =>
      val === undefined ||
      (/^\d+$/.test(val) && parseInt(val, 10) >= 1 && parseInt(val, 10) <= 1000),
    { message: "page must be an integer between 1 and 1000" }
  ),
```

Any `page` value outside 1–1000, or non-numeric, now returns **HTTP 400** before the controller is reached.

### Layer 2 — Controller clamp (`backend/controllers/matchController.js`)

```js
// BEFORE
const page = Math.max(1, parseInt(req.query.page, 10) || 1);
// AFTER
const page = Math.min(Math.max(1, parseInt(req.query.page, 10) || 1), 1000);
```

Maximum possible `.range()` offset: `(1000 - 1) × 100 = 99,900` rows — down from `(99999 - 1) × 100 = 9,999,800`.

## Changes

| File | Change |
|---|---|
| `backend/validation/schemas.js` | Add `page` field to `matchSchemas.getSupabaseDiscover` |
| `backend/controllers/matchController.js` | Clamp `page` to 1000 maximum |
| `backend/tests/supabaseDiscover.test.js` | New — 6 rejection cases, 2 boundary passes, 2 controller unit tests |

## Test coverage

**Schema rejection (400):** `page=99999` · `page=0` · `page=-1` · `page=1.5` · `page=abc` · `limit=101`

**Schema acceptance:** `page=1000` (boundary) · absent `page` param

**Controller unit tests:** `page=2&limit=10` → `skip=10, to=19` · `page=99999` clamped to max skip of `99900` (never reaches row 9,999,800)

## Release Notes

* **Bug Fixes**
  * Strengthened discover pagination handling by enforcing `page` within **1–1000** and `limit` within **1–100**, including clearer handling for out-of-range requests.

* **Tests**
  * Added Vitest coverage for the discover endpoint to verify validation errors for invalid `page`/`limit`, correct boundary behavior, and accurate pagination window calculations (including oversized pagination).


hariom888:
@durdana3105 These 6 CI failures are pre-existing on `main` and not introduced by this PR. The `express` module resolution error (`ERR_MODULE_NOT_FOUND`) and missing `docs/api.md` affect multiple test files that existed before this branch. All 75 previously passing tests continue to pass. This PR introduces no new failures.

durdana3105:
Please resolve the merge conflicts

PR #1015, opened 2026-06-16
fix(#951): extract rate-limiter factory to isolate cron and notification Map instances

### LINKED ISSUE #951: [Bug]: `backgroundRateCounts` and `isOnCooldown` Maps are shared across cron and notification route handlers — cross-endpoint DoS**
## Description

`backend/middlewares/requireCronSecret.js` exports both `isBackgroundRateLimited` (backed by `backgroundRateCounts`) and `isOnCooldown` (backed by `lastExecutions`) as module-level singletons.

Both `backend/routers/cronRoutes.js` (via `requireCronSecret`) and `backend/routers/notificationRoutes.js` (via the local `verifyNotificationAuth` function that explicitly imports `isBackgroundRateLimited` and `isOnCooldown`) consume the **same Map instances** from the same module import:

```js
// notificationRoutes.js lines 8-10
import {
  auditLog,
  isBackgroundRateLimited,  // ← same Map as used by cron routes
  isOnCooldown,
} from "../middlewares/requireCronSecret.js";
```

The shared rate limit window is `5 req/min per IP` (`BACKGROUND_MAX_REQUESTS = 5`).

## Reproduction

1. Send 5 rapid authenticated `POST /api/notifications/send-push` requests from a single IP using a valid `WEBHOOK_SECRET`.
2. Immediately attempt `POST /api/cron/dispatch-notifications` with a valid `CRON_SECRET` from the same IP.
3. The cron route is rejected with HTTP 429 — its budget has been consumed by the notification endpoint.

This is bidirectional: 5 cron calls also exhaust the budget for the notification endpoint.

## Impact

A Vercel Cron scheduler and an internal webhook service sharing the same egress IP (common in cloud deployments) will starve each other's rate-limit buckets. In the worst case, the cron job that drains the `notifications` push queue becomes permanently blocked during high-activity periods, causing notification delivery stalls that are invisible in logs (HTTP 429 is returned but not paged).

## Affected Files

- `backend/middlewares/requireCronSecret.js` — `backgroundRateCounts` and `lastExecutions` Maps, lines 11 and 35
- `backend/routers/notificationRoutes.js` — imports, lines 8–10; usage, lines 57 and 62

hariom888:
Closes #951

## Summary

`requireCronSecret.js` exported `isBackgroundRateLimited` and `isOnCooldown` as module-level singletons backed by a single `Map` each. Because Node.js caches modules, `notificationRoutes.js` importing those helpers received references to the exact same `Map` objects. Five rapid calls to `POST /api/notifications/send-push` from one IP consumed the entire 5 req/min budget, immediately 429-ing `POST /api/cron/dispatch-notifications` from the same IP — a common scenario when Vercel Cron and an internal webhook service share a cloud NAT egress address.

## Fix

Two factory functions replace the singleton exports:

```js
// New exports in requireCronSecret.js
export const createBackgroundRateLimiter = (windowMs, maxRequests) => { ... };
export const createCooldownTracker = (cooldownMs) => { ... };
```

Each caller creates its own instance:

```js
// requireCronSecret.js — cron routes
const cronRateLimiter = createBackgroundRateLimiter();
const cronCooldown    = createCooldownTracker();

// notificationRoutes.js — notification routes
const notificationRateLimiter = createBackgroundRateLimiter();
const notificationCooldown    = createCooldownTracker();
```

Each factory call allocates a fresh `Map`, so the two route families are fully isolated — exhausting one budget has zero effect on the other.

## Changes

| File | Change |
|---|---|
| `backend/middlewares/requireCronSecret.js` | Replace singleton exports with `createBackgroundRateLimiter` and `createCooldownTracker` factories; instantiate private cron instances |
| `backend/routers/notificationRoutes.js` | Import factories instead of singleton helpers; instantiate private notification instances |
| `backend/tests/rateLimiter.test.js` | New — 10 unit tests covering both factories including the isolation regression test |

## Test coverage

**Rate limiter:** first request allowed · up to max allowed · max+1 blocked · resets after window · different IPs independent · **two instances do not share state**

**Cooldown tracker:** first call allowed · blocked within window · resets after cooldown · different route keys independent · **two instances do not share state**

## Summary by CodeRabbit

* **Refactor**
  * Improved rate limiting and cooldown handling by isolating background job and webhook controls, reducing cross-request interference and improving reliability.
  * Strengthened webhook token verification using constant-time comparison, while keeping layered access checks and existing response behavior.
* **Tests**
  * Added/expanded test coverage for rate limiting and cooldown behavior, including per-IP/per-route enforcement and isolation between separately created limiters/trackers.

### DIFF
--- a/backend/middlewares/requireCronSecret.js
+++ b/backend/middlewares/requireCronSecret.js
@@ -8,43 +8,67 @@ import { HttpError } from "../utils/httpError.js";
  */
 const BACKGROUND_WINDOW_MS = 60_000;
 const BACKGROUND_MAX_REQUESTS = 5;
-const backgroundRateCounts = new Map();
 
-export const isBackgroundRateLimited = (ip) => {
-  const now = Date.now();
-  const entry = backgroundRateCounts.get(ip);
+/**
+ * Factory that returns an independent rate-limiter closure backed by its
+ * own Map instance. Callers (cron routes, notification routes) must each
+ * call createBackgroundRateLimiter() to get an isolated limiter so that
+ * one endpoint's traffic cannot exhaust another's budget.
+ *
+ * @param {number} windowMs   - Rolling window duration in ms (default 60 000)
+ * @param {number} maxRequests - Max allowed requests per window (default 5)
+ * @returns {(ip: string) => boolean}
+ */
+export const createBackgroundRateLimiter = (
+  windowMs = BACKGROUND_WINDOW_MS,
+  maxRequests = BACKGROUND_MAX_REQUESTS
+) => {
+  const counts = new Map();
 
-  if (!entry || now - entry.windowStart >= BACKGROUND_WINDOW_MS) {
-    backgroundRateCounts.set(ip, { count: 1, windowStart: now });
+  return (ip) => {
+    const now = Date.now();
+    const entry = counts.get(ip);
+
+    if (!entry || now - entry.windowStart >= windowMs) {
+      counts.set(ip, { count: 1, windowStart: now });
+      return false;
+    }
+
+    if (entry.count >= maxRequests) {
+      return true;
+    }
+
+    entry.count += 1;
     return false;
-  }
-
-  if (entry.count >= BACKGROUND_MAX_REQUESTS) {
-    return true;
-  }
-
-  entry.count += 1;
-  return false;
+  };
 };
 
 /**
- * Cooldown tracker: prevents re-invocation of expensive cron jobs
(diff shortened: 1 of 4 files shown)


hariom888:
@durdana3105 These 6 CI failures are pre-existing on `main` and not introduced by this PR. The `express` module resolution error (`ERR_MODULE_NOT_FOUND`) and missing `docs/api.md` affect multiple test files that existed before this branch. All 75 previously passing tests continue to pass. This PR introduces no new failures.

PR #1056, opened 2026-06-18
fix: replace fabricated pagination metadata with hasNextPage

### LINKED ISSUE #935: Bug: `getRecommendedPartners` Returns Intentionally Wrong `total` and `totalPages`, Pagination is Broken
### Description

The `/api/match/recommend` endpoint returns fabricated pagination metadata. The response fields `total` and `totalPages` are calculated using a hardcoded formula that always implies one more page exists, regardless of actual data. Any frontend component consuming these fields will loop indefinitely, always requesting a next page that comes back empty.

### Affected File

- `backend/controllers/matchController.js`

### Root Cause

The source code itself contains a comment acknowledging this is a deliberate shortcut:

```js
total: recommendations.length > 0 ? skip + limit + 1 : skip, // Rough pagination cursor hack
totalPages: recommendations.length === limit ? page + 1 : page, // always 1 more page
```

What this produces in practice:
- Page 1 returns 20 results → `totalPages` is set to `2`, even if only 20 users exist in total
- Page 2 returns 0 results → `totalPages` is still set to `2`, causing the UI to loop back on itself
- A "Load More" button or paginated list never reaches a genuine end state

### Impact

- Frontend pagination UI can loop infinitely, always believing there is a next page
- Any "Showing X of Y peers" counter will always display an incorrect total
- Poor UX for the Discover Peers feature, which is one of the platform's core flows

### Proposed Fix

There are two clean approaches. The first is to add a window function to the existing `match_users` Supabase RPC so that it returns the true total row count alongside each result row, the controller can then read this value from the first result and use it to compute `total` and `totalPages` correctly. The second, simpler option is to drop the `total` and `totalPages` fields entirely and replace them with a single `hasNextPage` boolean, which is set to `true` only when a full page of results was returned. This cursor-style approach requires no COUNT query at all and scales better as the user base grows. Either option should also update any frontend component currently consuming `total` or `totalPages` to match the new response shape.

Tamcodes4:
## Description

The recommendation endpoint was returning fabricated pagination metadata (`total` and `totalPages`) based on a heuristic that always implied another page could exist. This could cause pagination UIs to make unnecessary requests and display misleading counts.

This change replaces the fabricated metadata with a cursor-style `hasNextPage` flag.

### Changes

* Removed inaccurate `total` and `totalPages` response fields
* Fetches one extra record (`limit + 1`) to reliably determine whether another page exists
* Introduced `hasNextPage` boolean in the API response
* Trims the extra record before returning results to maintain the requested page size
* Preserves existing frontend behavior since consumers only use `recommendations`

### Why

The previous implementation used:

```js
total: recommendations.length > 0 ? skip + limit + 1 : skip
totalPages: recommendations.length === limit ? page + 1 : page
```

These values did not represent actual totals and could incorrectly indicate that another page existed even when the current page was the last one.

Using a `limit + 1` fetch provides a reliable and scalable way to determine pagination state without requiring a costly count query.

### Testing

* Verified recommendation consumers only use `recommendations`
* Confirmed no frontend usage of `total` or `totalPages`
* Verified `hasNextPage` is correctly set when additional results exist
* Verified response size remains capped to the requested page limit

Fixes #935 

## Release Notes

* **Bug Fixes**
  * Improved pagination for partner recommendations by accurately determining whether more results exist.
  * The response now returns a consistent number of recommendations per page and reports the availability of the next page reliably.
* **Tests**
  * Updated partner recommendation pagination tests to verify next-page detection and result trimming behavior.

### DIFF
--- a/backend/controllers/matchController.js
+++ b/backend/controllers/matchController.js
@@ -105,7 +105,7 @@ export const getRecommendedPartners = async (req, res) => {
       target_interests: currentUser.interests || [],
       target_teach: currentUser.teach_subjects || [],
       target_learn: currentUser.learn_subjects || [],
-      page_limit: limit,
+      page_limit: limit + 1,
       page_offset: skip
     });
 
@@ -133,13 +133,14 @@ export const getRecommendedPartners = async (req, res) => {
 
     // In a real paginated RPC, getting exact total Count requires a separate count query. 
     // We'll estimate or just provide length for now since counting 1M rows can also be slow.
+   const hasNextPage = recommendations.length > limit;
+
     res.status(200).json({
       success: true,
-      count: recommendations.length,
-      total: recommendations.length > 0 ? skip + limit + 1 : skip, // Rough pagination cursor hack
+      count: Math.min(recommendations.length, limit),
       page,
-      totalPages: recommendations.length === limit ? page + 1 : page,
-      recommendations,
+      hasNextPage,
+      recommendations: recommendations.slice(0, limit),
     });
   } catch (error) {
     console.error("Recommendation Error:", error);
--- a/backend/controllers/matchController.js
+++ b/backend/controllers/matchController.js
@@ -131,8 +131,8 @@ export const getRecommendedPartners = async (req, res) => {
       };
     });
 
-    // In a real paginated RPC, getting exact total Count requires a separate count query. 
-    // We'll estimate or just provide length for now since counting 1M rows can also be slow.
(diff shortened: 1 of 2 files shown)


Tamcodes4:
Hey @durdana3105
Bots commented issue has been resolved , kindly review this pr!

Tamcodes4:
Hey @durdana3105, I think you misinterpreted the labels 
This is clearly a bug fix....
Could you please change the docs label into bug?

PR #1125, opened 2026-06-20
Perf: Optimize React rendering and memoize heavy components (#989)

### LINKED ISSUE #989: [Bug/Frontend]: Optimize React Rendering in Real-Time and Heavy UI Components
Overview The application currently suffers from significant performance bottlenecks due to unoptimized React rendering cycles. Specifically, heavy UI components and those relying on real-time data streams (such as live chat, WebRTC rooms, and active timers) are frequently passing inline functions and unmemoized object literals as props to their child components.

Impact & Current State Because inline functions and raw objects receive a new reference identity on every single render cycle, React's reconciliation algorithm is forced to re-render all downstream child components—even if the actual data hasn't changed. In real-time components where state updates rapidly (e.g., ticking seconds in a Pomodoro timer, or receiving a stream of chat messages), this triggers massive, unnecessary render cascades across the DOM tree. This leads to severe UI lag, high CPU utilization on the client's device, degraded battery life on mobile, and an overall sluggish user experience during peak interactions.

Proposed Solution To stabilize performance and eliminate wasted render cycles, a targeted optimization sweep is required across the affected component trees:

Implement useCallback: Wrap all event handlers and functions passed as props to child components in useCallback hooks, ensuring their reference identities remain stable between renders.
Implement useMemo: Memoize expensive calculations, derived state variables, and object literal props using the useMemo hook.
Component Memoization (React.memo): Where appropriate, wrap heavy pure child components in React.memo() so they only re-render when their explicitly passed props change.
State Co-location: Review the affected files to ensure that rapidly changing state (like a ticking timer) is localized as deep in the component tree as possible, preventing it from unnecessarily updating parent wrappers.
Impacted Files (Non-Exhaustive List):

src/components/Chatbot.tsx
src/components/GroupPomodoro.tsx
src/components/Room.tsx
src/pages/Leaderboard.tsx
src/components/Navbar.tsx

ArshVermaGit:
This PR drastically improves client-side performance by surgically eliminating unnecessary React render cascades across our most interaction-heavy components. It implements stable reference identities by extracting inline functions into useCallback hooks across the Chatbot, Room, Leaderboard, and Navbar components, ensuring that deeply nested children are not forced to reconcile when parent wrapper states change. Furthermore, it hoists static utility functions (like formatTime in GroupPomodoro.tsx) out of component scopes to prevent continuous re-allocation during rapid real-time updates (like ticking timers). Finally, critical exports such as Navbar and GroupPomodoro have been shielded with React.memo(), guaranteeing they only re-render when explicitly required by strict prop equality checks. This suite of optimizations resolves UI lag and significantly drops CPU utilization during active study sessions. Closes #989


ArshVermaGit (commented):
resolved

durdana3105:
Please resolve merge conflicts

PR #1235, opened 2026-06-23
Fix/testimonials infinite scroll duplicate

### LINKED ISSUE #1192: fix(Testimonials): extract testimonials array and use spread for symmetric infinite scroll
Problem
In src/components/landing/Testimonials.tsx the infinite-scroll loop is implemented by duplicating testimonial objects inline in the JSX. Only 3 of the 6 unique cards are duplicated, meaning scrollWidth / 2 — the reset point — does not coincide with the true midpoint of the rendered list. The scroll resets at the wrong position, causing a visible jump on some viewport sizes. It also makes the component fragile: any future edit to the list requires manually synchronising the duplicates.

Root cause
The loop math assumes el.scrollWidth / 2 lands on a seamless repeat. This only holds if the second half of the list is an identical copy of the first half. With 6 + 3 items the assumption fails.

Changes

Extract the 6 unique testimonial objects into a const TESTIMONIALS array above the component (or into a separate testimonials.ts data file)
Replace the inline array in JSX with [...TESTIMONIALS, ...TESTIMONIALS] — this guarantees the two halves are identical and the scroll loop resets cleanly
Update the key prop: use key={`${t.name}-${i}`} (already present) — no change needed
Zero behaviour change for users; the loop now works correctly regardless of how many cards exist
PLEASE ASSIGN THE ISUUE TO ME UNDER GSSOC'26

OmanshiRaj:
fixes #1192 
In src/components/landing/Testimonials.tsx, the infinite-scroll carousel achieves its seamless loop by rendering the list twice and resetting scrollLeft when it reaches el.scrollWidth / 2 — but the duplication is done by manually hardcoding only 3 of the 6 unique testimonial objects after the originals, meaning the rendered list is 9 cards rather than 12, so scrollWidth / 2 lands in the middle of a card rather than at a clean boundary, causing a visible jump or stutter when the loop resets on certain viewport sizes and making the component fragile because any future addition or removal of a card requires manually keeping the inline duplicates in sync. This fix extracts all 6 unique testimonial objects into a TESTIMONIALS constant declared above the component and replaces the entire inline array literal in the JSX with [...TESTIMONIALS, ...TESTIMONIALS], which guarantees the two halves are always identical copies regardless of how many cards exist, makes the loop reset point mathematically correct, and reduces the JSX from 57 hardcoded lines down to a single spread expression — no new imports are needed and the card rendering markup inside the .map() callback is completely unchanged.

### DIFF
--- a/src/components/landing/Testimonials.tsx
+++ b/src/components/landing/Testimonials.tsx
@@ -4,6 +4,15 @@ import { Button } from "@/components/ui/button";
 import { useState, useEffect, useRef } from "react";
 import { toast } from "sonner";
 
+const TESTIMONIALS = [
+  { text: "PeerLearn helped me crack my first internship interview.", name: "Aisha", role: "AIML Student", rating: 5 },
+  { text: "I started mentoring juniors and improved my communication skills.", name: "Rahul", role: "Senior Mentor", rating: 5 },
+  { text: "Found amazing teammates for hackathons and projects.", name: "John", role: "Web Developer", rating: 4 },
+  { text: "Built a polished project portfolio with mentor guidance.", name: "Maya", role: "Frontend Developer", rating: 5 },
+  { text: "Mentors gave real-world advice that helped my internship prep.", name: "Priya", role: "ML Intern", rating: 5 },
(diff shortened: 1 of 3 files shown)


durdana3105:
please solve merge conflicts

durdana3105:

PR #1400, opened 2026-07-02
fix(sessions): sync selected session with filtered tabs

### LINKED ISSUE #1072: [Bug]: Switching session tabs keeps the old session selected in chat
\What is happening?  
When the user changes tabs, the session list changes, but the chat panel can still show the previously selected session.

Why is this a problem?  
Users may think they are viewing or chatting in a session from the selected tab, but the right-side chat may belong to a different session.

How to fix it:  
Update the selected session whenever `filteredSessions` or `selectedTab` changes. If the current selected session is not in the filtered list, select the first filtered session or clear it.

Steps to Reproduce:  
1. Open the Sessions page.  
2. Select a session from one tab.  
3. Switch to another tab.  
4. Notice the chat panel may still show the old session.

Files afftected:  
`src/hooks/useSessions.ts`

mayurigade-hub:
## Related Issue
Closes #1072

## Description

Fixed an issue where the chat panel continued displaying a previously selected session after switching session tabs.

Previously, changing the selected tab updated the session list but did not update the selected session. As a result, the chat panel could continue showing a session that no longer belonged to the active tab, leading to inconsistent behavior.

## Changes Made

- Updated the selected session whenever `filteredSessions` or `selectedTab` changes
- Checked whether the current selected session exists in the filtered session list
- Automatically selected the first available session when the previous selection is no longer valid
- Cleared the selected session when the filtered list is empty
- Kept the session list and chat panel synchronized

## Steps to Test

1. Open the Sessions page
2. Select a session from one tab
3. Switch to another tab
4. Verify the chat panel updates to a valid session in the selected tab
5. Verify the first available session is selected when appropriate
6. Verify the chat panel clears if there are no sessions in the selected tab

## Expected Result

- The selected session always belongs to the active tab.
- The chat panel stays synchronized with the filtered session list.
- No stale session remains selected after switching tabs.

## Files Changed

- `src/hooks/useSessions.ts`

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session filtering so tabs now map consistently to the correct session statuses.
  * Kept the selected session in sync with the active filter: if a session is no longer visible, the first matching session is selected automatically, or the selection is cleared when none remain.


mayurigade-hub:
Hello @durdana3105 
Please review this PR. 
Thank you!

PR #1469, opened 2026-07-02
Include backend in root workspace

### LINKED ISSUE #1458: Root install does not prepare backend test dependencies
## Summary
A root-only install leaves backend test imports unresolved. The full test suite only passes after installing dependencies inside the backend package as well.

## Steps to reproduce
1. Clone the repository.
2. Run npm install from the repository root.
3. Run npm test -- --run from the repository root.

## Actual result
Backend test suites fail to import packages such as express and supertest.

## Expected result
The documented root setup should prepare everything needed for the root test command, or the README should clearly call out the required backend install step.

## Suggested fix
Either convert the repo to npm workspaces so the root install includes backend dependencies, or document and script the two-step install/test workflow.

Tomeshwari-02:
## Summary
- Adds the backend package as a root npm workspace
- Updates the root lockfile so backend dependencies are installed by the root install flow

## Validation
- npm install
- npm test -- --run

Closes #1458

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,9 @@
     "": {
       "name": "vite_react_shadcn_ts",
       "version": "0.0.0",
+      "workspaces": [
+        "backend"
+      ],
       "dependencies": {
         "@hookform/resolvers": "^3.10.0",
         "@jitsi/react-sdk": "^1.4.4",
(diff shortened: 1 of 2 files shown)


durdana3105:
Hi @Tomeshwari-02! 👋

Your PR includes the backend as a root npm workspace, which is great! However, the CI linter job is currently failing with:

```
sh: 1: eslint: not found
```

**The Issue:**
The workflow runs `npm run lint` (which calls ESLint), but ESLint isn't available because dev dependencies aren't being installed in the CI environment.

**The Fix:**
Update `.github/workflows/ci.yml` line 24 to explicitly include dev dependencies:

```yaml
- name: Install Dependencies
  run: npm ci --legacy-peer-deps --include=dev
```

This ensures ESLint and other dev dependencies are properly installed before the linter runs.

**Why This Happens:**
Your changes introduce npm workspaces, and with the current CI config, dev dependencies at the root level may not be installed. By adding `--include=dev`, we guarantee all dev dependencies are available.

Once you make this change, the CI should pass! Let me know if you have any questions. 🚀

---

Tomeshwari-02:
Hii @durdana3105, Thanks for the feedback! 🙌 I've implemented the suggested changes and pushed the fix. The CI checks are now passing. Could you please take another look when you have a chance? Thank you!

durdana3105:

PR #1531, opened 2026-07-03
test: add automated RLS coverage check for every frontend-queried table

### LINKED ISSUE #1525: Security: Supabase anon key in Vite bundle without Row Level Security on all tables

anshul23102:
## Summary

Fixes #1525

## Context

I investigated this issue and found RLS is already enabled on every table the frontend queries — see the migration history (`20260403063915_...`, `20260518_role_management.sql`, `20260617000000_consolidate_rls_policies.sql`, and 15+ other targeted RLS-hardening migrations). There's no live gap matching the issue as filed today.

What *was* missing: nothing prevented a future PR from adding a new `supabase.from("new_table")` call in the frontend without a matching RLS migration, silently reopening exactly the class of bug this issue describes.

## Changes

- `src/test/rlsCoverage.test.ts`: a static-analysis test (no live DB connection needed) that:
  1. Scans `src/**/*.{ts,tsx}` for `.from("table_name")` calls
  2. Scans `supabase/migrations/*.sql` for `ALTER TABLE ... ENABLE ROW LEVEL SECURITY` statements
  3. Fails, listing the exact table names, if any frontend-queried table has no matching RLS migration

Currently passes: all 23 tables queried from the frontend have RLS enabled. This becomes a permanent regression guard going forward.

## Testing

- `npx vitest run src/test/rlsCoverage.test.ts` — 2/2 pass
- `npm run lint` — 0 errors

### DIFF
--- a/src/test/rlsCoverage.test.ts
+++ b/src/test/rlsCoverage.test.ts
@@ -1,0 +1,100 @@
+import { describe, it, expect } from "vitest";
+import { readFileSync, readdirSync } from "fs";
+import { resolve } from "path";
+
+// Regression coverage for #1525: the frontend embeds the Supabase anon key
+// in the production bundle, which is fine *only* if every table reachable
+// through `supabase.from(...)` has Row Level Security enabled. RLS is
+// already turned on for every table currently queried from the frontend
+// (see the migrations under supabase/migrations/), but there was no
+// automated check preventing a future PR from adding a new
+// `supabase.from("some_new_table")` call without a matching
+// `ALTER TABLE ... ENABLE ROW LEVEL SECURITY` migration. This test closes
+// that gap statically, without needing a live database in CI.
+
+const ROOT = resolve(__dirname, "../..");
+const SRC_DIR = resolve(ROOT, "src");
+const MIGRATIONS_DIR = resolve(ROOT, "supabase/migrations");
+
+// Tables that are intentionally public / have no user-owned rows, or that
+// are views backed by an RLS-protected base table. Extend this list with a
+// comment explaining why whenever a genuinely public table is added.
+const RLS_EXEMPT_TABLES = new Set<string>([]);
+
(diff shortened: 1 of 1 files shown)


anshul23102:

Suggested labels for this PR:
- `type:security`
- `level:beginner`
- `gssoc26`
- `gssoc:approved`

PR #1532, opened 2026-07-03
test: add end-to-end coverage proving /api/chat requires authentication

### LINKED ISSUE #1526: Security: OpenRouter API key forwarded by backend without request validation or auth

anshul23102:
## Summary

Fixes #1526

## Context

I investigated this issue and found `/api/chat` already requires authentication — `requireAuth` is applied ahead of the handler in `backend/routers/chatRoutes.js`, and it correctly verifies the Supabase JWT (local HMAC verification, rejecting missing/invalid/expired tokens with 401). There's no live bypass matching the issue as filed today.

What *was* missing: `requireAuth` had unit tests in isolation (`requireAuth.test.js`) but nothing exercised the actual `POST /api/chat` route end-to-end to prove the real Express app wires the middleware in correctly.

## Changes

- `backend/tests/chatRoute.auth.integration.test.js`: hits the real app via supertest, covering:
  - no `Authorization` header -> 401, OpenRouter never called
  - malformed token -> 401
  - validly-formed but wrongly-signed token -> 401
  - correctly-signed token -> 200, OpenRouter called once

This locks in the fix as a regression guard: a future refactor that accidentally drops `requireAuth` from the middleware chain will fail this test instead of silently reopening the vulnerability.

## Testing

- `npx vitest run backend/tests/chatRoute.auth.integration.test.js` — 4/4 pass
- `npm run lint` — 0 errors

### DIFF
--- a/backend/tests/chatRoute.auth.integration.test.js
+++ b/backend/tests/chatRoute.auth.integration.test.js
@@ -1,0 +1,123 @@
+import crypto from "crypto";
+import request from "supertest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// Regression coverage for #1526: POST /api/chat previously had no
+// integration test proving requireAuth actually gates the route before a
+// request reaches OpenRouter. requireAuth itself is already unit-tested in
+// isolation (requireAuth.test.js), but nothing exercised the real Express
+// app to confirm chatRoutes.js wires it in correctly.
+
+const JWT_SECRET = "chat-auth-route-test-secret";
+
+const base64UrlEncode = (value) =>
+  Buffer.from(JSON.stringify(value))
+    .toString("base64")
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_")
+    .replace(/=/g, "");
+
+const createLocalJwt = (payload, secret = JWT_SECRET) => {
+  const header = base64UrlEncode({ alg: "HS256", typ: "JWT" });
+  const body = base64UrlEncode(payload);
+  const signature = crypto
+    .createHmac("sha256", secret)
+    .update(`${header}.${body}`)
+    .digest("base64")
+    .replace(/\+/g, "-")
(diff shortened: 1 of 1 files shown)


anshul23102:

Suggested labels for this PR:
- `type:security`
- `level:beginner`
- `gssoc26`
- `gssoc:approved`

PR #1533, opened 2026-07-03
test: add end-to-end coverage proving /api/chat enforces per-user rate limiting

### LINKED ISSUE #1528: Bug: POST /api/chat has no per-user rate limiting, single user can exhaust API quota

anshul23102:
## Summary

Fixes #1528

## Context

I investigated this issue and found `/api/chat` already has per-user rate limiting — `rateLimiter` (`backend/middlewares/rateLimiter.js`) is applied ahead of the handler, keys on `req.user.id` when authenticated, caps at 20 requests/minute, and returns 429 with standard `X-RateLimit-*` headers once exceeded. There's no live gap matching the issue as filed today.

What *was* missing: `rateLimiter` had unit tests in isolation (`rateLimiter.test.js`) but nothing exercised the actual `POST /api/chat` route end-to-end to prove the real Express app wires it in correctly ahead of the OpenRouter call.

## Changes

- `backend/tests/chatRoute.rateLimit.integration.test.js`: hits the real app via supertest, covering:
  - 25 rapid requests from one authenticated user -> 20 succeed, 5 throttled with 429, and OpenRouter is called exactly 20 times (never for the excess)
  - the 429 response includes a clear error message and the documented `X-RateLimit-*` headers
  - two different users' quotas are independent of each other

This locks in the fix as a regression guard: a future refactor that accidentally drops the rate limiter from the middleware chain will fail this test instead of silently reopening the quota-drain vulnerability.

## Testing

- `npx vitest run backend/tests/chatRoute.rateLimit.integration.test.js` — 3/3 pass
- `npm run lint` — 0 errors

### DIFF
--- a/backend/tests/chatRoute.rateLimit.integration.test.js
+++ b/backend/tests/chatRoute.rateLimit.integration.test.js
@@ -1,0 +1,149 @@
+import crypto from "crypto";
+import request from "supertest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// Regression coverage for #1528: POST /api/chat previously had no
+// integration test proving the rate limiter actually caps a single user's
+// throughput on the real route. rateLimiter is already unit-tested in
+// isolation (rateLimiter.test.js), but nothing exercised the real Express
+// app to confirm chatRoutes.js wires it in correctly ahead of the
+// OpenRouter call.
+
+const JWT_SECRET = "chat-ratelimit-route-test-secret";
+
+const base64UrlEncode = (value) =>
+  Buffer.from(JSON.stringify(value))
+    .toString("base64")
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_")
+    .replace(/=/g, "");
+
+const createLocalJwt = (payload, secret = JWT_SECRET) => {
+  const header = base64UrlEncode({ alg: "HS256", typ: "JWT" });
+  const body = base64UrlEncode(payload);
+  const signature = crypto
+    .createHmac("sha256", secret)
+    .update(`${header}.${body}`)
+    .digest("base64")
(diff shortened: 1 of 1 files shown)


anshul23102:

Suggested labels for this PR:
- `type:security`
- `level:beginner`
- `gssoc26`
- `gssoc:approved`

PR #1534, opened 2026-07-03
feat: add secure-by-default private storage helpers

### LINKED ISSUE #1529: Security: session replay recordings stored in public Supabase bucket without access control

anshul23102:
## Summary

Fixes #1529

## Context

I investigated this issue and found that this repo doesn't currently implement session replay recordings at all — there's no bucket, upload path, or feature anywhere in `src` or `backend` matching what's described. The issue is filed against a feature that doesn't exist yet.

Its underlying concern is legitimate though: Supabase Storage buckets default to public unless explicitly created with `public: false`, so whenever a recording (or any other private-file) feature is built here, it's one missed flag away from being world-readable — exactly the failure mode the issue describes.

## Changes

- `backend/utils/privateStorage.js`:
  - `ensurePrivateBucket(name)`: idempotently creates a bucket with `public: false`. If a bucket of that name already exists and is public, it warns rather than silently flipping it — changing visibility on an existing bucket can break already-issued public links and should be a deliberate migration, not an automatic side effect.
  - `getSignedFileUrl(bucket, path, expiresInSeconds)`: wraps `storage.from(bucket).createSignedUrl(...)` with input validation and a normalized `HttpError` on failure.

Neither function is wired into a route yet, since there's no replay feature to attach them to. This is the secure foundation described in the issue's own proposed fix (private bucket + signed URLs), ready for whoever builds session replay recording to use from day one instead of hitting the Supabase default-public trap.

## Testing

- `npx vitest run backend/tests/privateStorage.test.js` — 9/9 pass
- `npm run lint` — 0 errors

### DIFF
--- a/backend/tests/privateStorage.test.js
+++ b/backend/tests/privateStorage.test.js
@@ -1,0 +1,127 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { HttpError } from "../utils/httpError.js";
+
+const mockGetBucket = vi.fn();
+const mockCreateBucket = vi.fn();
+const mockCreateSignedUrl = vi.fn();
+const mockFrom = vi.fn(() => ({ createSignedUrl: mockCreateSignedUrl }));
+
+vi.mock("../utils/supabase.js", () => ({
+  getSupabaseAdmin: () => ({
+    storage: {
+      getBucket: mockGetBucket,
+      createBucket: mockCreateBucket,
+      from: mockFrom,
+    },
+  }),
+}));
+
+const { ensurePrivateBucket, getSignedFileUrl } = await import("../utils/privateStorage.js");
+
+describe("ensurePrivateBucket (#1529)", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("creates a new bucket with public:false when it does not exist", async () => {
+    mockGetBucket.mockResolvedValue({
+      data: null,
+      error: { message: "Bucket not found" },
+    });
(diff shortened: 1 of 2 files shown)


anshul23102:

Suggested labels for this PR:
- `type:feature`
- `level:intermediate`
- `gssoc26`
- `gssoc:approved`

PR #1557, opened 2026-07-04
Restrict resource writes to owners

### LINKED ISSUE #1547: Security: consolidated RLS allows any user to modify or delete any resource
## Problem
The consolidated RLS migration redefines resource policies with unrestricted checks:

```sql
CREATE POLICY "Users can insert resources"
  ON public.resources FOR INSERT WITH CHECK (true);
CREATE POLICY "Users can update resources"
  ON public.resources FOR UPDATE USING (true);
CREATE POLICY "Users can delete resources"
  ON public.resources FOR DELETE USING (true);
```

This appears to override the earlier resource ownership model from `20260518000000_resource_hub.sql`, where inserts were tied to `auth.uid() = uploaded_by` and deletes were restricted to the uploader.

## Impact
Any client using the Supabase anon key can potentially insert arbitrary resource rows, edit existing resource metadata, or delete resources uploaded by other users, depending on table grants. This breaks ownership guarantees in the resource hub.

## Affected file
- `supabase/migrations/20260617000000_consolidate_rls_policies.sql` lines 423-431

## Expected behavior
Resource writes should be limited to authenticated users acting on their own rows. Suggested policy shape:

```sql
FOR INSERT WITH CHECK (auth.uid() = uploaded_by)
FOR UPDATE USING (auth.uid() = uploaded_by) WITH CHECK (auth.uid() = uploaded_by)
FOR DELETE USING (auth.uid() = uploaded_by)
```

The exact column name should match the final `resources` schema.

Tomeshwari-02:
## Problem
The consolidated RLS migration redefines resource policies with unrestricted checks:

```sql
CREATE POLICY "Users can insert resources"
  ON public.resources FOR INSERT WITH CHECK (true);
CREATE POLICY "Users can update resources"
  ON public.resources FOR UPDATE USING (true);
CREATE POLICY "Users can delete resources"
  ON public.resources FOR DELETE USING (true);
```

This appears to override the earlier resource ownership model from `20260518000000_resource_hub.sql`, where inserts were tied to `auth.uid() = uploaded_by` and deletes were restricted to the uploader.

## Fixes 
Restores owner-based RLS checks for resource inserts, updates, and deletes while keeping public read access. Validation: SQL-only policy change reviewed against the existing resources ownership column used by the app. 

Fixes #1547

### DIFF
--- a/supabase/migrations/20260617000000_consolidate_rls_policies.sql
+++ b/supabase/migrations/20260617000000_consolidate_rls_policies.sql
@@ -420,15 +420,15 @@ CREATE POLICY "Users can delete own push subscriptions"
 --------------------------------------------------------------------------------
 -- 10. RESOURCES & WHITEBOARD
 --------------------------------------------------------------------------------
--- resources (Assuming existence)
+-- resources
 CREATE POLICY "Anyone can view resources" 
   ON public.resources FOR SELECT USING (true);
 CREATE POLICY "Users can insert resources" 
-  ON public.resources FOR INSERT WITH CHECK (true);
+  ON public.resources FOR INSERT WITH CHECK (uploaded_by = auth.uid());
 CREATE POLICY "Users can update resources" 
-  ON public.resources FOR UPDATE USING (true);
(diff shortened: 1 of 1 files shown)


durdana3105:

PR #1609, opened 2026-07-04
Constrain upload storage paths

### LINKED ISSUE #1599: Upload endpoint allows users to choose arbitrary storage paths
﻿### Summary
The upload endpoint trusts `folder` and `filePath` from form data. An authenticated user can choose any allowed bucket and any object path instead of being constrained to their own namespace.

### Where
- `backend/controllers/uploadController.js` reads `folder` and `filePath` from `req.body`.
- The handler only checks whether the bucket name is in the allow-list.

### Steps to reproduce
1. Authenticate and submit an upload request.
2. Set `folder` to an allowed bucket and set `filePath` to another user's prefix or an unrelated path.
3. The backend uploads using the supplied path with the service client.

### Expected behavior
The server should derive the destination path from the authenticated user and upload type, or verify that the supplied path belongs to the current user before using the service client.

Tomeshwari-02:
﻿## Summary
- reject upload paths outside the signed-in user's storage prefix
- reject empty, traversal, or unsafe path segments before using the service client

## Validation
- npm test -- --runInBand backend/tests/uploadPhoto.test.js

Closes #1599

### DIFF
--- a/backend/controllers/uploadController.js
+++ b/backend/controllers/uploadController.js
@@ -6,6 +6,16 @@ import { HttpError } from "../utils/httpError.js";
 
 // Ensure files do not exceed 50MB
 const MAX_FILE_SIZE = 50 * 1024 * 1024;
+const SAFE_PATH_SEGMENT = /^[a-zA-Z0-9._-]+$/;
+
+const isUserScopedPath = (filePath, userId) => {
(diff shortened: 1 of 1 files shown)


durdana3105:

PR #1616, opened 2026-07-04
Validate session seat limits

### LINKED ISSUE #1605: Session creation accepts invalid seat limits from the form
﻿### Summary
The create-session form treats `seatLimit` as an optional string and parses it during submit without validating that it is a positive integer. Invalid strings, zero, or negative values can reach the insert payload.

### Where
- `src/hooks/useCreateSession.ts` defines `seatLimit` as `z.string().optional()`.
- Submit logic uses `parseInt(values.seatLimit, 10)` and sends the result as `seat_limit`.
- The migration that adds `seat_limit` does not add a positive-value check.

### Steps to reproduce
1. Enter a non-numeric, zero, or negative seat limit in the create-session form.
2. Submit the form.
3. The parsed value is passed into the database insert instead of being rejected by form validation.

### Expected behavior
The form schema should validate seat limits as positive integers before submit, and the database should also enforce the same invariant.

Tomeshwari-02:
﻿## Summary
- validate optional session seat limits as positive whole numbers in the form schema
- parse only the trimmed validated value during submit
- add a database check constraint for persisted seat limits

## Validation
- npm run typecheck (fails on existing unrelated errors in docs, resources, and generated Supabase table types)

Closes #1605

### DIFF
--- a/src/hooks/useCreateSession.ts
+++ b/src/hooks/useCreateSession.ts
@@ -18,7 +18,16 @@ export const formSchema = z
     time: z.string().min(1, "Time is required."),
     durationPreset: z.number().optional(),
     durationCustom: z.string().optional(),
-    seatLimit: z.string().optional(),
+    seatLimit: z
+      .string()
+      .optional()
+      .refine(
+        (value) => {
+          const trimmed = value?.trim();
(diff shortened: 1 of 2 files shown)


durdana3105:

Issue #1614, opened 2026-07-04
[Bug]: Validation error message doesn't clear after fixing the issue (Skills & Experience steps)


Sumit-programmer-787:
## Bug Summary
On the "Become a Mentor" onboarding flow, when a validation error message appears (e.g. "Please select at least one skill" or missing GitHub/LinkedIn links), fixing the underlying issue does not clear the error message from the screen. The message only disappears after clicking "Continue" a second time — it doesn't clear reactively as the user corrects their input.

## Steps to Reproduce
1. Visit https://peer-learning-liart.vercel.app/become-mentor
2. Go to the **Skills** step
3. Click "Continue" without selecting any skill → error shows: "Please select at least one skill"
4. Now select or add a skill
5. Observe: the error message is still visible on screen, even though the form is now valid

Same behavior reproduces on the **Experience** step:
1. Go to the Experience step
2. Click "Continue" with GitHub/LinkedIn fields empty → error shows: "Please fill GitHub and LinkedIn profiles"
3. Fill in both fields
4. Observe: the error message stays visible until Continue is clicked again

## Expected Behavior
Once the underlying validation issue is resolved (e.g. a skill is selected, or the required fields are filled), the error message should clear automatically and immediately — without requiring another click on "Continue".

## Actual Behavior
The error message persists on screen even after the user has corrected their input. It only clears once "Continue" is clicked again and validation passes (at which point the user is also navigated to the next step, so there's no visible "success" feedback moment — it just disappears as the step changes).

## Environment
- Website: https://peer-learning-liart.vercel.app/become-mentor
- Browser: Google Chrome (latest)
- OS: Windows 11

## Impact
- Confusing UX: users may think their input still isn't valid, or repeatedly try to fix something that's already fixed, since the error message gives no live feedback.
- Affects at least two steps of the mentor onboarding form (Skills, Experience); likely a form-wide pattern rather than an isolated case.

## Root Cause (from initial investigation)
In `src/components/mentor/MentorForm.tsx`, the `error` state is only cleared inside the "Continue" button's `onClick` handler, and only once all validations pass:
```tsx
setError("");
setStep(step + 1);
```
No other handler (`toggleSkill`, `handleAddCustomSkill`, `onChange` on the Experience inputs, etc.) ever resets `error`, so it stays stale on screen until the next full validation pass.

## Proposed Solution
[...]

Issue #1604, opened 2026-07-04
Room chat records activity before send succeeds


Tomeshwari-02:
﻿### Summary
Room chat adds "You sent a message" to the activity feed before the database insert finishes. The Room wrapper also adds the same activity after calling the send handler, so successful sends can create duplicate entries and failed sends can still appear as sent.

### Where
- `src/hooks/useRoomChat.ts` adds activity before inserting into `study_room_messages`.
- `src/components/Room/Room.tsx` adds another activity after `handleSendMessage` resolves.

### Steps to reproduce
1. Open a study room and send a message.
2. Watch the activity feed after a successful send.
3. Force the insert to fail and send again.

### Expected behavior
The activity feed should add a single success entry only after the insert succeeds, and should avoid logging sent activity when the insert fails.

PR #1631, opened 2026-07-04
fix: address six bugs (#1614, #1599, #1598, #1605, #1604, #1596)

saurabhhhcodes:
### #1614 — Validation error message doesn't clear after fixing input
Added setError("") to every input's onChange handler so validation errors clear reactively when the user corrects their input.

### #1599 — Upload endpoint allows arbitrary storage paths
Added path traversal sanitization to filePath param using path.normalize() and rejection of paths containing ..

### #1598 — Chat completion accepts client system messages
Filter out any messages with role: "system" from client payload before constructing the completion request, preventing prompt injection.

### #1605 — Session creation accepts invalid seat limits
Added Zod validation for seatLimit field: must be a positive integer between 1 and 1000.

### #1604 — Room chat records activity before send succeeds
Moved setActivities call to after the Supabase insert succeeds, and added return on failure to prevent phantom activity entries.

### #1596 — Avatar upload accepts SVG client-side but backend rejects
Explicitly reject image/svg+xml MIME type in the upload controller file filter.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upload safety by rejecting unsafe file paths and blocking SVG image uploads.
  * Prevented client-provided system messages from affecting chat requests.
  * Fixed message activity logging so failed sends no longer appear as successful.
  * Clear form errors as soon as users edit affected profile fields.
  * Tightened session seat-limit validation to accept only values from 1 to 1000.

### DIFF
--- a/backend/controllers/uploadController.js
+++ b/backend/controllers/uploadController.js
@@ -1,6 +1,7 @@
 import multer from "multer";
 import os from "os";
 import fs from "fs";
+import path from "path";
 import { getSupabaseAdmin } from "../utils/supabase.js";
 import { HttpError } from "../utils/httpError.js";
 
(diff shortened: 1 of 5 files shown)


durdana3105:
PLEASE RESOLVE MERGE CONFLICTS AND CODERABBIT COMMENTS

PR #1683, opened 2026-07-05
fix(discover): roll back optimistic connect state and toast on failure

### LINKED ISSUE #1672: Bug: Discover Connect button stays Pending forever when the request fails (no optimistic rollback)
### Description

**Bug Summary**

On the Discover page, clicking "Connect" optimistically marks the peer as connected, which disables the button and shows "Pending". If the `peer_connections` insert fails, the code does not roll back the optimistic state and shows no error, so the button stays disabled at "Pending" forever and the user can never retry.

**Root Cause**

`src/pages/Discover.tsx`:

```jsx
setConnections((prev) => [...prev, peerId]);
const { error } = await supabase.from("peer_connections").insert({...});
if (!error) {
  await supabase.from("notifications").insert({...});
}
// no else: on error, connections still contains peerId
```

The button is `disabled={isConnected}` and renders "Pending" when `connections.includes(u.id)`, so a failed insert leaves it permanently stuck. The sibling `Dashboard.handleConnect` rolls back on error, but this page was missed.

**Code Location**

- `src/pages/Discover.tsx` (`handleConnect`)

### Steps to Reproduce

1. Open Discover.
2. Cause the `peer_connections` insert to fail (for example, offline, or an RLS rejection).
3. Click "Connect" on a peer.
4. The button switches to "Pending" and stays disabled forever, with no error shown.

### Expected Behavior

If the connection request fails, the optimistic state is rolled back so the button returns to "Connect", the user can retry, and an error is shown.

### Actual Behavior

The button stays disabled at "Pending" permanently, with no feedback.

### Impact

- A single failed request permanently disables the Connect button for that peer, with no way to retry and no error message.

### Proposed Fix

On error, remove the peer from the optimistic `connections` list and show a toast, matching the rollback the Dashboard already does:

```js
if (error) {
  setConnections((prev) => prev.filter((id) => id !== peerId));
  toast.error("Could not send the connection request. Please try again.");
  return;
}
```

### Verification

- The fix mirrors the rollback pattern already used in `Dashboard.handleConnect`.
- No render test was added: simulating a failing Supabase insert inside the Discover page is disproportionate to a rollback/toast fix.

### Suggested labels

bug, ui/ux

Anexus5919:
## Summary

On Discover, "Connect" optimistically set the peer as connected (disabling the button to "Pending"), but a failed `peer_connections` insert never rolled that back and showed no error, so the button stayed stuck at "Pending". This rolls back and toasts on failure, matching `Dashboard.handleConnect`.

## Changes

- `Discover.tsx`: on insert error, remove the peer from the optimistic `connections` list and show a `toast.error`, then return; success path is unchanged.

## Verification

- Mirrors the rollback already used in `Dashboard.handleConnect`.
- No render test was added: simulating a failing Supabase insert inside the Discover page is disproportionate to a rollback/toast fix.

## Related

Fixes #1672

### DIFF
--- a/src/pages/Discover.tsx
+++ b/src/pages/Discover.tsx
@@ -11,6 +11,7 @@ import {
   Zap,
 } from "lucide-react";
 
+import { toast } from "sonner";
 import { supabase } from "@/integrations/supabase/client";
 import { NotificationsDropdown } from "@/components/NotificationsDropdown";
 import { Check } from "lucide-react";
@@ -277,14 +278,18 @@ const Discover = () => {
       status: 'pending'
     });
 
-    if (!error) {
-      await (supabase as any).from("notifications").insert({
-        user_id: peerId,
-        type: 'system',
-        title: 'New Connection Request',
-        body: `${currentUser.name || 'Someone'} wants to connect with you!`,
-      });
+    if (error) {
+      setConnections((prev) => prev.filter((id) => id !== peerId));
+      toast.error("Could not send the connection request. Please try again.");
+      return;
     }
+
+    await (supabase as any).from("notifications").insert({
+      user_id: peerId,
+      type: 'system',
+      title: 'New Connection Request',
+      body: `${currentUser.name || 'Someone'} wants to connect with you!`,
+    });
   }, [connections, currentUser]);
 
   return (


Anexus5919:

PR #1684, opened 2026-07-05
fix(backend): apply the 50kb AI body limit shadowed by the global parser

### LINKED ISSUE #1657: Bug: the 50KB AI-route body limit is dead code, shadowed by the global 100KB parser
### Description

**Bug Summary**

The AI routes are meant to enforce a tighter 50KB request body limit than the global 100KB limit, but that 50KB cap never applies. The global `express.json({ limit: "100kb" })` middleware runs first and parses the body, so the per-route `express.json({ limit: "50kb" })` sees an already parsed body and returns immediately without enforcing its own limit.

**Root Cause**

`body-parser` marks the request with `req._body = true` after the first successful parse. Every later `express.json()` invocation checks that flag and calls `next()` without re-reading or re-limiting the body.

In `backend/app.js` the global parser is mounted before any route:

```js
app.use(express.json({ limit: "100kb" }));
```

and `backend/routers/aiRoutes.js` mounts a second parser per AI route:

```js
const aiBodyLimit = express.json({ limit: "50kb" });
router.post("/ask", aiBodyLimit, requireAuth, ...);
```

Because the AI router is mounted after the global parser (`app.use("/api/ai", aiRoutes)`), the 100KB parser always runs first, so `aiBodyLimit` is inert. The documented AI specific limit does not exist in practice.

**Code Location**

- `backend/app.js` (global `express.json` at 100KB)
- `backend/routers/aiRoutes.js` (per-route 50KB parser, shadowed)

### Steps to Reproduce

1. Send `POST /api/ai/ask` with `Content-Type: application/json` and a body between 50KB and 100KB (for example an object with an 80KB string).
2. Observe the request is accepted by the parser (it proceeds to authentication) instead of being rejected with 413 at 50KB.

### Expected Behavior

AI routes reject request bodies larger than 50KB with 413, while other routes keep the 100KB limit.

### Actual Behavior

AI routes accept bodies up to the global 100KB limit; the 50KB cap is never enforced.

### Impact

- The intended tighter limit for AI endpoints, which forward to a paid model, is not in effect, so a larger than intended payload is accepted.

### Proposed Fix

Mount the 50KB parser for `/api/ai` before the global parser so it runs first, and drop the now redundant per-route parser:

```js
app.use("/api/ai", express.json({ limit: "50kb" }));
app.use(express.json({ limit: "100kb" }));
```

### Verification

Reproduced and fixed against a real Express instance and the actual `app.js`:

- Before: an 80KB body to `/api/ai/ask` returns 200 (accepted).
- After: an 80KB body returns 413; a small body returns 401 (parsed, reaches authentication); a non-AI route still accepts up to 100KB.
- Added a backend test (`backend/tests/aiBodyLimit.test.js`); the full backend suite passes (140 tests).

### Suggested labels

bug, backend, security

Anexus5919:
## Summary

The AI routes' 50KB body limit never applied. The global `express.json({ limit: "100kb" })` parses the body first, so the per-route 50KB parser was inert (body-parser's `req._body` short-circuit).

## Changes

- `app.js`: mount a 50KB json parser for `/api/ai` before the global 100KB parser, so it runs first.
- `aiRoutes.js`: remove the now redundant per-route parser.
- Add `tests/aiBodyLimit.test.js`.

## Verification

Against a real Express instance and the actual `app.js`:

- Before: an 80KB body to `/api/ai/ask` returned 200.
- After: 80KB returns 413; a small body returns 401 (parsed, reaches auth); non-AI routes still accept up to 100KB.
- Full backend suite passes (140 tests).

## Related

Fixes #1657

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -66,6 +66,8 @@ app.use(cors({
   allowedHeaders: ['Content-Type', 'Authorization']
 }));
 
+// AI routes use a tighter body limit; mount before the global parser so it applies.
+app.use("/api/ai", express.json({ limit: "50kb" }));
 app.use(express.json({ limit: "100kb" }));
 app.use(cookieParser());
 app.use((req, res, next) => {
--- a/backend/routers/aiRoutes.js
+++ b/backend/routers/aiRoutes.js
@@ -15,13 +15,9 @@ import { aiSchemas } from "../validation/schemas.js";
 
 const router = express.Router();
 
-// SECURITY: Tighter body limit for AI routes (50KB) since these endpoints
-// have well-defined, smaller input requirements than the global 100KB cap.
-const aiBodyLimit = express.json({ limit: "50kb" });
-
-router.post("/ask", aiBodyLimit, requireAuth, rateLimiter, validate(aiSchemas.askAI), asyncHandler(askAI));
-router.post("/generate-summary", aiBodyLimit, requireAuth, rateLimiter, validate(aiSchemas.generateSessionSummary), asyncHandler(generateSessionSummary));
-router.post("/mock-interview/chat", aiBodyLimit, requireAuth, rateLimiter, validate(aiSchemas.mockInterviewChat), asyncHandler(conductMockInterview));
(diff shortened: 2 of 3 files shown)


Anexus5919:

PR #1685, opened 2026-07-05
fix(backend): stop logging intentional HttpErrors as unhandled errors

### LINKED ISSUE #1661: Bug: global error handler logs every error as Unhandled error, double-logging and mislabeling intentional HttpErrors
### Description

**Bug Summary**

The global Express error handler has an unconditional `console.error("Unhandled error:", err)` that runs before it classifies the error. As a result, every intentional `HttpError` (for example a 401 from `requireAuth`) is first logged as an "Unhandled error", then logged again with its real label. Unknown errors are also logged twice, and expected client errors are recorded at error level.

**Root Cause**

In `backend/middlewares/errorHandler.js`:

```js
console.error("Unhandled error:", err);           // runs for every non-Zod/non-Multer error
// --- Known HttpError ---
if (err instanceof HttpError) {
  const status = err.statusCode || 500;
  console.error(`[${requestId}] HttpError ${status}:`, err.message);
  ...
}
// --- Unknown ---
console.error(`[${requestId}] Unhandled error:`, err);
```

The unconditional line runs after the Zod and Multer early-returns but before the HttpError branch, so:

- An `HttpError` is logged at the unconditional line as "Unhandled error", then at the HttpError branch as "HttpError <status>".
- An unknown error is logged at the unconditional line and again at the unknown branch.

**Code Location**

- `backend/middlewares/errorHandler.js`

### Steps to Reproduce

1. Send a request without a token to any route protected by `requireAuth`.
2. Check the server logs.
3. Observe an "Unhandled error: ..." line for the expected 401, followed by the intended "HttpError 401: ..." line.

### Expected Behavior

An intentional `HttpError` is logged once, at an appropriate level (warn for 4xx, error for 5xx). An unknown error is logged once as unhandled.

### Actual Behavior

Every error is logged twice, and expected 4xx client errors are recorded as unhandled server errors.

### Impact

- Log noise: every request that produces an `HttpError` writes two lines.
- Misleading alerting and forensics: expected 401/403/404 responses appear as unhandled server errors.

### Proposed Fix

Remove the unconditional log line and rely on the branch-specific logs. Log expected 4xx HttpErrors at warn and 5xx at error:

```js
if (err instanceof HttpError) {
  const status = err.statusCode || 500;
  const log = status >= 500 ? console.error : console.warn;
  log(`[${requestId}] HttpError ${status}:`, err.message);
  ...
}
```

### Verification

- Added a backend test asserting: a 401 HttpError is not logged as "Unhandled error" and is logged at warn; an unknown error is logged exactly once as unhandled; a 5xx HttpError logs at error. The test fails on the old code and passes on the fix.
- Full backend suite passes.

### Suggested labels

bug, backend

Anexus5919:
## Summary

The global error handler had an unconditional `console.error("Unhandled error:", err)` before it classified the error, so every intentional `HttpError` was logged twice, and expected 4xx client errors were recorded as unhandled server errors. This removes the stray line and logs 4xx at warn, 5xx at error.

## Changes

- `errorHandler.js`: remove the unconditional log; log HttpError at warn (status < 500) or error (>= 500).
- Add `tests/errorHandler.test.js`.

## Verification

- New test: a 401 HttpError is not logged as "Unhandled error" and is logged at warn; an unknown error is logged exactly once as unhandled; a 5xx HttpError logs at error. The test fails on the old code and passes on the fix.
- Full backend suite passes.

## Related

Fixes #1661

### DIFF
--- a/backend/middlewares/errorHandler.js
+++ b/backend/middlewares/errorHandler.js
@@ -61,11 +61,11 @@ export const errorHandler = (err, req, res, next) => {
     return res.status(400).json({ error: err.message });
   }
 
-  console.error("Unhandled error:", err);
   // --- Known HttpError (intentionally thrown by our code) ---
   if (err instanceof HttpError) {
     const status = err.statusCode || 500;
-    console.error(`[${requestId}] HttpError ${status}:`, err.message);
+    const log = status >= 500 ? console.error : console.warn;
+    log(`[${requestId}] HttpError ${status}:`, err.message);
 
     const payload = { error: err.message };
 
--- a/backend/tests/errorHandler.test.js
+++ b/backend/tests/errorHandler.test.js
@@ -1,0 +1,50 @@
+import { describe, it, expect, vi, afterEach } from "vitest";
+import { errorHandler } from "../middlewares/errorHandler.js";
+import { HttpError } from "../utils/httpError.js";
+
+const makeRes = () => ({
+  headersSent: false,
+  statusCode: null,
+  body: null,
+  status(code) { this.statusCode = code; return this; },
+  json(payload) { this.body = payload; return this; },
+  setHeader() {},
+});
+
+describe("errorHandler logging", () => {
+  afterEach(() => vi.restoreAllMocks());
+
+  it("does not log an intentional 4xx HttpError as an unhandled error", () => {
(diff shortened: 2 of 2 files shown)


Anexus5919:

Issue #1663, opened 2026-07-05
Bug: whiteboard strokes use raw pixel coordinates, so drawings land in the wrong place for participants with different canvas sizes


Anexus5919:
### Description

**Bug Summary**

The collaborative whiteboard captures, broadcasts, and persists stroke points as raw device pixels relative to each client's own canvas. Because each client sizes the canvas to its own container, two participants whose window or panel sizes differ see strokes at different positions, and persisted history is misaligned for anyone whose canvas differs from the author's.

**Root Cause**

In `src/components/Whiteboard/Canvas.tsx`:

- The canvas is sized to the local container:

```js
canvas.width = container.clientWidth;
canvas.height = container.clientHeight;
```

- Points are captured as raw pixels:

```js
return { x: e.clientX - rect.left, y: e.clientY - rect.top };
```

- Those raw pixels are broadcast and persisted verbatim, and replayed verbatim:

```js
ctx.moveTo(point.x, point.y); // draw-start
ctx.lineTo(point.x, point.y); // draw-move
```

There is no normalization by canvas size anywhere, so a point at x=800 on a 900px canvas is drawn at x=800 on a 500px canvas.

**Code Location**

- `src/components/Whiteboard/Canvas.tsx` (`resizeCanvas`, `getCoordinates`, `drawEvent`)

### Steps to Reproduce

1. Open a room as User A with the whiteboard maximized (canvas about 900px wide).
2. Open the same room as User B with the participant panel visible (canvas about 500px wide).
3. As A, draw a stroke near the right edge (x about 800).
4. On B's canvas the stroke renders at x=800, off the right side or clipped, not where A drew it.

### Expected Behavior

A stroke drawn at a given relative position on one canvas renders at the same relative position on every other participant's canvas, regardless of size.

### Actual Behavior

Strokes render at absolute pixel positions, so they are offset or clipped for anyone whose canvas differs from the author's.

### Impact

- The core promise of a shared whiteboard (everyone sees the same board) is broken whenever two participants have different canvas sizes, which is the normal case.

### Proposed Fix

Store points as fractions of the canvas size (0..1) and denormalize on draw using the local canvas dimensions. To stay compatible with already-persisted raw-pixel events, tag new events with a `normalized` flag and only denormalize those; legacy events keep drawing verbatim.

### Verification

[...]

PR #1688, opened 2026-07-05
fix(whiteboard): isolate strokes per id so concurrent drawing does not corrupt the canvas

### LINKED ISSUE #1664: Bug: concurrent whiteboard drawing corrupts the canvas because a remote draw-start resets the shared path mid-stroke
### Description

**Bug Summary**

`drawEvent` relies on the single shared 2D canvas path and keys only on the event type. `strokeId` is captured on every event but ignored during drawing. When a remote `draw-start` arrives while a local stroke is in progress, its `ctx.beginPath(); ctx.moveTo(...)` resets the shared path, so the next local `draw-move` draws a `lineTo` from the remote start point, producing a stray line across the board. Any two participants drawing at the same time corrupt each other's strokes, and history replay of interleaved strokes is affected too.

**Root Cause**

`src/components/Whiteboard/Canvas.tsx`, `drawEvent`:

```js
if (event.type === "draw-start") { ctx.beginPath(); ctx.moveTo(point.x, point.y); }
if (event.type === "draw-move")  { ctx.lineTo(point.x, point.y); ctx.stroke(); }
```

Local `pushEvent` and the remote broadcast handler both feed into this one shared path with no per-stroke isolation, so interleaved events collide. `strokeId` is present in the payload but never used here.

**Code Location**

- `src/components/Whiteboard/Canvas.tsx` (`drawEvent`)

### Steps to Reproduce

1. Open the same room in two clients.
2. Both users draw at the same time.
3. Observe stray diagonal lines that belong to neither user's stroke.

Concretely: A is mid-stroke at (20,20). B's `draw-start` (200,200) arrives and resets the shared path. A's next `draw-move` (30,30) draws a line from (200,200) to (30,30).

### Expected Behavior

Each participant's stroke is drawn independently; concurrent drawing never connects points across different strokes.

### Actual Behavior

A remote `draw-start` resets the shared path mid-stroke, so the next local move draws a spurious segment from the other stroke's point.

### Impact

- Concurrent drawing (the expected collaboration scenario) produces stray lines, and interleaved history replays incorrectly.

### Proposed Fix

Track the last point per `strokeId` and draw each `draw-move` as a discrete segment (beginPath, moveTo(prev), lineTo(current), stroke) using the stroke's own previous point, instead of relying on the shared canvas path state. Round line caps make the discrete segments visually identical to a continuous stroke.

### Verification

- Added a unit test of the per-stroke segment logic with interleaved strokes A and B: after B's `draw-start` arrives mid-stroke, A's next move connects from A's own last point, not B's start. A first move with no prior start draws a dot; `draw-end` clears the stroke's state.
- The visual multi-client rendering is not covered by an automated test, since jsdom does not render canvas. The segment logic and wiring are covered by the unit test and typecheck.

### Suggested labels

bug, whiteboard

Anexus5919:
## Summary

`drawEvent` relied on the single shared canvas path, so a remote `draw-start` arriving mid-stroke reset the path and the next local `draw-move` drew a stray line from the other stroke's point. This isolates strokes per `strokeId` and draws discrete segments.

## Changes

- `strokePath.ts`: pure `nextSegment` helper that tracks the last point per `strokeId` and returns the segment to draw.
- `Canvas.tsx`: `drawEvent` draws isolated per-stroke segments; reset the per-stroke state on replay and clear.
- `strokePath.test.ts`: unit tests for interleaved strokes.

Round line caps keep the discrete segments visually identical to a continuous stroke.

## Verification

- Unit test: with strokes A and B interleaved, A's move after B's `draw-start` connects from A's own last point (20,20), not B's start (200,200). A first move with no start draws a dot; `draw-end` clears the stroke.
- The visual multi-client rendering is not covered by an automated test (jsdom does not render canvas); the segment logic and wiring are covered by the unit test and typecheck.

## Note

This touches the same `drawEvent` function as the coordinate-normalization fix (#1663 / branch `fix/whiteboard-coordinate-normalization`). If both are merged, a small rebase is needed; the two changes compose cleanly (normalize the point, then feed it through `nextSegment`).

## Related

Fixes #1664

### DIFF
--- a/src/components/Whiteboard/Canvas.tsx
+++ b/src/components/Whiteboard/Canvas.tsx
@@ -3,9 +3,11 @@ import { useEffect, useRef, useState } from "react";
 import { supabase } from "@/integrations/supabase/client";
 import { useAuth } from "@/contexts/useAuth";
 import {
+  Point,
   ToolType,
   WhiteboardEvent,
 } from "./types";
+import { nextSegment } from "./strokePath";
 
 type Props = {
   roomId: string;
@@ -22,6 +24,8 @@ export default function Canvas({ roomId }: Props) {
   const strokesRef = useRef<WhiteboardEvent[]>([]);
 
   const currentStrokeId = useRef<string | null>(null);
+
+  const lastPointRef = useRef<Map<string, Point>>(new Map());
 
   const { user } = useAuth();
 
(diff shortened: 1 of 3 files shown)


Anexus5919:

PR #1194, opened 2026-06-22
[GSSoC'26] fix: eliminate duplicate database query in Messages

vipul674:
## Description

- Replaced duplicate `Promise.all` with a single Supabase query to the `profiles` table
- Profiles are guaranteed to exist for every authenticated user (created in `AuthContext.ensureProfileExists`), so the separate `users` table query was redundant
- Removed the now-unused `mergeProfiles` function (dead code)
- Reduces database round-trips from 2 to 1 per Messages page load

## Files Changed

- `src/pages/Messages.tsx`: Replaced Promise.all(profiles + users) with single profiles query; removed mergeProfiles function

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

## Difficulty & Label Request

Assessed difficulty: `level2`

Please apply the matching difficulty label for GSSoC '26 scoring.
If applicable, please also apply `gssoc:approved` after review.

## Testing & Verification

Commands run:

```bash
npm run build
```

Result: Build succeeds.

## Known Limitations

- The `Chat.tsx` page has a similar two-query pattern but was not in scope for this issue. It can be addressed separately if needed.

## GSSoC 2026 Compliance

This contribution was prepared with AI assistance and manually reviewed before submission.

Closes #1180

Issue #1180, opened 2026-06-22
[Bug]: useMessages hook fires two identical profile queries, wasting database resources


vipul674:
## Summary

The `useMessages` hook in `src/hooks/useMessages.ts` fires two identical Supabase queries to the `profiles` table with the same filters inside a `Promise.all`. Both queries select all columns, filter out the current user, order by name, and limit to 100 rows. This is clearly redundant and wastes database and network resources.

## Evidence

In `src/hooks/useMessages.ts`, lines 184-198:

```typescript
const [{ data: profileData, error: profileError }, { data: userData, error: userError }] =
  await Promise.all([
    supabase
      .from("profiles")
      .select("*")
      .neq("id", currentUserId)
      .order("name", { ascending: true })
      .limit(100),
    supabase
      .from("profiles")
      .select("*")
      .neq("id", currentUserId)
      .order("name", { ascending: true })
      .limit(100),
  ]);
```

Both queries are identical:
- Same table: `profiles`
- Same select: `*`
- Same filter: `.neq("id", currentUserId)`
- Same order: `.order("name", { ascending: true })`
- Same limit: `.limit(100)`

The `mergeProfiles` function on line 200 then merges two identical arrays, producing the same result as a single query would.

## Affected files

- `src/hooks/useMessages.ts` (lines 184-203)

## Reproduction steps

1. Open the Messages page.
2. Open browser DevTools Network tab.
3. Observe two identical `profiles` requests being fired simultaneously.
4. Both return the same data.

## Expected behavior

A single query should be used to fetch profiles. The `mergeProfiles` call with two identical arrays is unnecessary.

## Actual behavior

Two identical database queries are fired, doubling the network and database load for this operation.

## Impact

- Doubles the database load for every Messages page visit.
- Doubles the network bandwidth consumed.
- Increases latency unnecessarily (two round-trips instead of one).
- The `mergeProfiles` function produces the same result whether given one or two identical arrays.

## Suggested fix

Replace the `Promise.all` with a single query:

```typescript
const { data: profileData, error: profileError } = await supabase
  .from("profiles")
  .select("*")
  .neq("id", currentUserId)
  .order("name", { ascending: true })
  .limit(100);

const mergedUsers = (profileData ?? []).map((row) => row as ProfileSummary);
```

## GSSoC 2026

I would like to work on this issue under GSSoC 2026. Please assign it to me.

vipul674:
[...]

PR #1689, opened 2026-07-05
perf(chat): remove duplicate profiles query on the chat user list

### LINKED ISSUE #1665: Bug: /chat page fires two byte-identical profiles queries on every load
### Description

**Bug Summary**

The `/chat` page loads its user list with `Promise.all` of two byte-identical queries against the `profiles` table, then merges and dedupes them. The second query is pure waste: it doubles the database round-trips for the user list on every load and returns exactly the same rows.

**Root Cause**

`src/pages/Chat.tsx`:

```js
const [{ data: profileData }, { data: userData }] = await Promise.all([
  supabase.from("profiles").select("*").neq("id", currentUser.id).order("name", { ascending: true }).limit(100),
  supabase.from("profiles").select("*").neq("id", currentUser.id).order("name", { ascending: true }).limit(100), // identical
]);
const mergedUsers = mergeUsers((profileData ?? []) as Profile[], (userData ?? []) as UserRow[]);
```

Both queries hit `profiles` with the same filters, and `mergeUsers` dedupes by id, so the second query and the merge add nothing. `UserRow` is just an alias of `Profile`, which suggests the second query was a copy-paste that was never pointed at a different source.

**Code Location**

- `src/pages/Chat.tsx` (`loadUsers`)

### Steps to Reproduce

1. Open the `/chat` page while signed in.
2. Inspect the network tab (or Supabase logs).
3. Observe two identical requests to the `profiles` table for the same user list on every load.

### Expected Behavior

The chat user list is fetched with a single query.

### Actual Behavior

Two identical `profiles` queries run on every load, doubling the round-trips.

### Impact

- Doubles the database load for the chat user list on every visit to `/chat`.

### Proposed Fix

Query `profiles` once and use the result directly, removing the now-unnecessary merge. This matches how the same pattern was fixed for the Messages hook.

### Verification

- Behavior is unchanged: the merge only deduplicated by id, which a single primary-key-keyed query already guarantees.

### Note

This is the same class of bug that was fixed for the `useMessages` hook in #1180 / PR #1194, but this second instance in `src/pages/Chat.tsx` was not addressed by that change.

### Suggested labels

bug, performance, good first issue

Anexus5919:
## Summary

The `/chat` page fetched its user list with two byte-identical `profiles` queries via `Promise.all`, then merged and deduped them. This queries `profiles` once and removes the now-unnecessary merge, matching how the same pattern was fixed for the Messages hook (#1180 / PR #1194).

## Changes

- `Chat.tsx`: single `profiles` query for the user list; remove the redundant `mergeUsers` helper and the `UserRow` alias.

## Verification

- Behavior is unchanged: the merge only deduplicated by id, which a single primary-key-keyed query already guarantees.

## Related

Fixes #1665

### DIFF
--- a/src/pages/Chat.tsx
+++ b/src/pages/Chat.tsx
@@ -18,22 +18,6 @@ type Profile = {
   avatar_url?: string | null;
 };
 
-type UserRow = Profile;
-
-const mergeUsers = (profiles: Profile[], users: UserRow[]) => {
-  const map = new Map<string, Profile>();
-
-  for (const user of users) {
-    map.set(user.id, user);
-  }
-
-  for (const profile of profiles) {
-    map.set(profile.id, profile);
-  }
-
-  return Array.from(map.values());
-};
-
 type ChatMessage = {
   id: string;
   sender_id: string | null;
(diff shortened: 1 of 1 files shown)


Anexus5919:

PR #1690, opened 2026-07-05
fix(sessions): clear messages when switching sessions

### LINKED ISSUE #1666: Bug: session chat shows the previous session's messages after switching (messages not reset)
### Description

**Bug Summary**

When the user switches the selected study session, the chat panel keeps showing the previous session's messages until the new session's message fetch resolves. The per-session reset effect clears the participant count and the AI summary, but not the messages, so the old thread lingers for the whole network round-trip.

**Root Cause**

`src/hooks/useSessions.ts`:

```js
useEffect(() => {
  setParticipantCount(1);
  setSessionSummary(null);
}, [selectedSession]);
```

This resets `participantCount` and `sessionSummary` but not `messages`. The messages fetch is awaited before `setMessages(data)` runs:

```js
const { data } = await supabase.from("messages").select("*").eq("session_id", selectedSession.id) ...;
setMessages(data || []);
```

So between switching sessions and the fetch resolving, the previous session's messages remain in state and rendered.

**Code Location**

- `src/hooks/useSessions.ts` (per-session reset effect)

### Steps to Reproduce

1. Open the Sessions page and select session A that has messages.
2. Select session B.
3. Observe A's messages remain visible until B's messages finish loading. This is obvious on a slow connection or when B has no messages (A's thread stays until the empty result arrives).

### Expected Behavior

Switching sessions immediately clears the chat, so the previous session's messages never show under the new session.

### Actual Behavior

The previous session's messages linger for the duration of the new session's fetch.

### Impact

- The user briefly sees the wrong conversation, and could read or reply against a stale thread during the fetch window.

### Proposed Fix

Clear `messages` in the same per-session reset effect that already clears `participantCount` and `sessionSummary`:

```js
useEffect(() => {
  setParticipantCount(1);
  setSessionSummary(null);
  setMessages([]);
}, [selectedSession]);
```

### Verification

- The fix mirrors the sibling resets already in this effect for participant count and AI summary.

### Note

This is the same class of stale-state-on-switch bug as the participant count (#1402) and AI summary (#1401), which were fixed by resetting those values here; the identical gap for `messages` was overlooked. Distinct from #1488, which is about the messages fetch silently discarding errors.

### Suggested labels

bug, frontend

Anexus5919:
## Summary

Switching the selected session kept the previous session's messages rendered until the new fetch resolved, because the per-session reset effect cleared `participantCount` and `sessionSummary` but not `messages`. This clears `messages` in the same effect.

## Changes

- `useSessions.ts`: add `setMessages([])` to the per-session reset effect.

## Verification

- The fix mirrors the sibling resets already in this effect for participant count and AI summary.

## Related

Fixes #1666

### DIFF
--- a/src/hooks/useSessions.ts
+++ b/src/hooks/useSessions.ts
@@ -96,9 +96,12 @@ export function useSessions(user: any) {
   // - sessionSummary: without this, a summary generated for a previous
   //   video session can keep showing after switching to a different
   //   session, making it look like it belongs to the new one.
+  // - messages: without this, the previous session's thread stays rendered
+  //   for the whole fetch round-trip after switching.
   useEffect(() => {
     setParticipantCount(1);
     setSessionSummary(null);
+    setMessages([]);
   }, [selectedSession]);
 
   useEffect(() => {


Anexus5919:

PR #1691, opened 2026-07-05
fix(mentor): enforce mentorship-type selection before submitting the application

### LINKED ISSUE #1667: Bug: MentorForm lets you submit with zero mentorship types because the validation is unreachable
### Description

**Bug Summary**

The mentor application form never enforces the mentorship-type selection. The `validateMentorship()` guard is placed inside the Continue button, which only renders while `step < 3`. On the final step (`step === 3`) the Submit button is shown instead and calls `handleSubmit` directly, so a user can reach the last step, select no mentorship type, and submit an application with `mentorship_types: []`.

**Root Cause**

`src/components/mentor/MentorForm.tsx`:

- The Continue button (rendered only when `step < 3`) contains a check that can never run:

```js
if (step === 3 && !validateMentorship()) { ... }  // step is always < 3 here
```

- The Submit button (rendered when `step === 3`) calls `handleSubmit` with no validation:

```js
<button type="button" onClick={handleSubmit} disabled={loading}>Submit Application</button>
```

- `handleSubmit` inserts into `mentors` without checking `validateMentorship()`.

**Code Location**

- `src/components/mentor/MentorForm.tsx` (Continue button dead branch, and `handleSubmit`)

### Steps to Reproduce

1. Open the Become a Mentor form.
2. Fill steps 1 to 3 with valid data.
3. On the Mentorship step, select no mentorship type.
4. Click Submit Application.
5. The application is inserted with `mentorship_types: []`.

### Expected Behavior

Submitting requires at least one mentorship type; otherwise an error is shown and no application is created.

### Actual Behavior

The application submits with an empty mentorship-type list, because the validation only exists in an unreachable branch of the Continue button.

### Impact

- Mentor applications can be created with no mentorship types, which the form intends to require.

### Proposed Fix

Enforce the check where the submit actually happens by early-returning in `handleSubmit` when no type is selected, and remove the unreachable branch from the Continue button:

```js
const handleSubmit = async () => {
  if (!validateMentorship()) {
    setError("Please select at least one mentorship type");
    return;
  }
  ...
};
```

### Verification

- The fix reuses the existing error message and matches the per-step validation pattern used for the other steps.
- No automated component test was added: the form has no existing tests and a behavioral test would require driving all four steps and mocking the Supabase auth, skills, and mentors calls, which is disproportionate to a validation-guard fix.

### Suggested labels

bug, ui/ux

Anexus5919:
## Summary

The mentor application form never enforced the mentorship-type selection: the `validateMentorship()` check lived in an unreachable branch of the Continue button (only rendered when `step < 3`), while the Submit button on the last step called `handleSubmit` directly. This enforces the check in `handleSubmit` and removes the dead branch.

## Changes

- `MentorForm.tsx`: early-return in `handleSubmit` when no mentorship type is selected (reusing the existing error message); remove the unreachable `step === 3` branch from the Continue button.

## Verification

- The fix matches the per-step validation pattern used for the other steps.
- No component test was added: the form has no existing tests and a behavioral test would require driving all four steps and mocking the Supabase auth, skills, and mentors calls, which is disproportionate to a validation-guard fix.

## Related

Fixes #1667

### DIFF
--- a/src/components/mentor/MentorForm.tsx
+++ b/src/components/mentor/MentorForm.tsx
@@ -103,6 +103,10 @@ export default function MentorForm() {
     return formData.mentorship_types.length > 0;
   };
   const handleSubmit = async () => {
+    if (!validateMentorship()) {
+      setError("Please select at least one mentorship type");
+      return;
+    }
     setLoading(true);
     setError("");
     let isTimeout = false;
@@ -402,10 +406,6 @@ export default function MentorForm() {
                 }
                 if (step === 2 && !validateExperience()) {
                   setError("Please fill GitHub and LinkedIn profiles");
-                  return;
-                }
-                if (step === 3 && !validateMentorship()) {
-                  setError("Please select at least one mentorship type");
                   return;
                 }
                 setError("");


Anexus5919:

PR #1724, opened 2026-07-08
fix(security): drop User-Agent from rate-limit key to prevent bucket rotation

### LINKED ISSUE #1659: Security: rate-limit key includes the User-Agent header, so the per-IP limit is bypassable by rotating it
### Description

**Bug Summary**

The in-memory rate limiter derives its per-request key from the client IP and the User-Agent header for unauthenticated requests. Because the User-Agent is fully attacker-controlled, sending each request with a different User-Agent produces a new bucket (count starts at 1), so the limit never triggers for unauthenticated traffic. It also inflates the number of distinct keys, which accelerates FIFO eviction of legitimate entries.

**Root Cause**

`backend/middlewares/rateLimiter.js` builds the key as:

```js
const ua = req.headers["user-agent"] || "no-ua";
return `ip:${expressIp}|${socketIp}|${ua}`;
```

The User-Agent is part of the primary bucket discriminator. The comment claims the socket IP "still anchors them to their real origin", but since the key is a concatenation, a different User-Agent yields an entirely different key regardless of IP.

This applies to unauthenticated routes that use this limiter (for example the pre-auth `POST /api/auth/set-cookie`). Authenticated requests are keyed by user id and are not affected. The global express-rate-limit on `/api/*` partly mitigates the impact.

**Code Location**

- `backend/middlewares/rateLimiter.js` (`deriveRateLimitKey`)

### Steps to Reproduce

1. Send more than the limit of requests (default 20 per minute) to a route protected by this limiter from a single IP.
2. Set a different `User-Agent` header on each request.
3. Observe no `429` is returned by this limiter, because each User-Agent gets its own bucket.

### Expected Behavior

Unauthenticated requests from the same client IP share one bucket, so the limit applies regardless of the User-Agent.

### Actual Behavior

Each distinct User-Agent from the same IP is a separate bucket, so the limit is trivially bypassed.

### Impact

- The custom per-IP rate limit can be bypassed by header rotation on unauthenticated routes.
- Many distinct keys speed up eviction of legitimate entries at the `maxEntries` cap.

### Proposed Fix

Key unauthenticated traffic on the client IP only, never on attacker-controlled headers:

```js
return `ip:${expressIp}|${socketIp}`;
```

### Verification

- Added a backend test: three requests then a fourth, all from the same IP with different User-Agent values. The fourth returns 429 with the fix and returns 200 (bypass) without it.
- Full backend suite passes.

### Suggested labels

bug, security, backend

Anexus5919:
## Summary

The unauthenticated rate-limit key included the User-Agent header, so rotating that header produced a fresh bucket on every request and bypassed the per-IP limit. This keys unauthenticated traffic on the client IP only.

## Changes

- `rateLimiter.js`: drop the User-Agent from `deriveRateLimitKey`; key on `req.ip` plus the raw socket remote address. Corrected the stale comment.
- Add `tests/rateLimiterKey.test.js`.

## Verification

- New test: three requests then a fourth, all from the same IP with different User-Agent values. The fourth returns 429 with the fix and returns 200 (bypass) without it.
- Full backend suite passes.

## Related

Fixes #1659

### DIFF
--- a/backend/middlewares/rateLimiter.js
+++ b/backend/middlewares/rateLimiter.js
@@ -20,11 +20,12 @@ const CLEANUP_INTERVAL_MS = 60 * 1000;
  * Derives a rate-limit key for the current request.
  *
  * Priority:
- *   1. Authenticated user ID (most reliable - cannot be spoofed).
- *   2. Composite fingerprint combining req.ip, the raw socket remote address,
- *      and the User-Agent header. This ensures that even if an attacker spoofs
- *      X-Forwarded-For (when trust proxy is misconfigured), the raw socket IP
- *      still anchors them to their real origin.
+ *   1. Authenticated user ID (cannot be spoofed).
+ *   2. Client IP: req.ip plus the raw socket remote address, so a spoofed
+ *      X-Forwarded-For still resolves to the real socket origin.
+ *
+ * Client-controlled headers such as User-Agent are excluded, since an attacker
+ * could rotate them to obtain fresh buckets and bypass the limit.
  */
 const deriveRateLimitKey = (req) => {
   if (req.user?.id) {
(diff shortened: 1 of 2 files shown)


Anexus5919:

PR #1725, opened 2026-07-08
fix(backend): correct unintended character range in email validation regex

### LINKED ISSUE #1660: Bug: email validation regex has an unintended character range that accepts junk local parts and rejects hyphens
### Description

**Bug Summary**

The email validation regex in `sendEmail.js` uses the character class `[a-z0-9._%-+]`. The `%-+` sequence is parsed as a range from `%` (0x25) to `+` (0x2B), not as three literals. As a result the local-part matcher accepts characters it should reject (`&`, `'`, `(`, `)`, `*`), and because the hyphen is consumed as the range operator, it also rejects a literal hyphen that the code intends to allow.

**Root Cause**

In `backend/utils/sendEmail.js`:

```js
const emailRegex = /^[a-z0-9._%-+]+@[a-z0-9.-]+\.[a-z]{2,}$/;
```

Inside a character class, `%-+` is a range (0x25 to 0x2B) covering `% & ' ( ) * +`. The author intended the three literals `%`, `+`, and `-` (per the comment: "alphanumeric, dots, hyphens, underscores, plus signs"). The literal hyphen (0x2D) falls outside the accidental range, so it does not match.

**Code Location**

- `backend/utils/sendEmail.js` (`emailRegex`)

### Steps to Reproduce

- `isValidEmail("a'()*@example.com")` returns `true` (should be `false`).
- `isValidEmail("a-b@example.com")` returns `false` (should be `true`, since a hyphen is allowed).

### Expected Behavior

The local part accepts letters, digits, and the intended specials `. _ % + -`, and rejects other punctuation.

### Actual Behavior

The local part accepts `& ' ( ) *` and rejects a literal hyphen.

### Impact

- Addresses with unintended characters pass validation.
- Valid addresses with a hyphen in the local part are rejected.

### Proposed Fix

Reorder the class so the hyphen is the last member and is treated as a literal:

```js
const emailRegex = /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$/;
```

### Verification

- Compared the old and new regex directly: the new class matches the intended set and rejects the junk characters.
- Added a backend test (`emailValidation.test.js`) covering valid addresses (including a hyphen local part), rejected junk characters, and missing-TLD cases. The test fails on the old regex and passes on the fix. The full backend suite passes.

### Suggested labels

bug, backend, good first issue

Anexus5919:
## Summary

The email validation regex used the character class `[a-z0-9._%-+]`, where `%-+` is an unintended range (`%` to `+`). It accepted `& ' ( ) *` and rejected a literal hyphen in the local part. Reordering the hyphen to the end makes `%`, `+`, and `-` literals.

## Changes

- `sendEmail.js`: change the class to `[a-z0-9._%+-]`; export `isValidEmail` for testing.
- Add `tests/emailValidation.test.js`.

## Verification

- New test covers valid addresses (including a hyphen local part), rejected junk characters, and missing-TLD cases. It fails on the old regex and passes on the fix.
- Full backend suite passes.

## Related

Fixes #1660

### DIFF
--- a/backend/tests/emailValidation.test.js
+++ b/backend/tests/emailValidation.test.js
@@ -1,0 +1,32 @@
+import { describe, it, expect } from "vitest";
+import { isValidEmail } from "../utils/sendEmail.js";
+
+describe("isValidEmail", () => {
+  it("accepts valid addresses, including a hyphen in the local part", () => {
+    for (const email of [
+      "user@example.com",
+      "user.name@example.com",
+      "user+tag@example.com",
+      "user_name@example.com",
+      "a-b@example.com",
+      "first%last@example.com",
+    ]) {
+      expect(isValidEmail(email)).toBe(true);
+    }
+  });
+
+  it("rejects characters outside the intended local-part set", () => {
+    for (const email of [
+      "a'()*@example.com",
+      "a b@example.com",
+      "a<b@example.com",
+    ]) {
+      expect(isValidEmail(email)).toBe(false);
+    }
+  });
+
+  it("rejects addresses without a valid TLD", () => {
+    expect(isValidEmail("a@b")).toBe(false);
+    expect(isValidEmail("user@example")).toBe(false);
+  });
+});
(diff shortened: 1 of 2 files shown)


Anexus5919:

PR #1783, opened 2026-07-16
fix(db): remove no-op migration that disabled award_activity_xp

### LINKED ISSUE #1633: Bug: migration 20260611000001 overwrites award_activity_xp with an empty no-op body, disabling all activity XP
### Description

**Bug Summary**

The migration `supabase/migrations/20260611000001_fix_xp_forgery.sql` replaces the working `award_activity_xp(text)` function with an empty body. After it runs the function still exists and is executable, but it does nothing: no `user_activity_log` insert and no `increment_user_xp` call. All activity XP silently stops.

**Root Cause**

The file contains a single statement that overwrites the real implementation:

```sql
CREATE OR REPLACE FUNCTION award_activity_xp(_activity_type TEXT) RETURNS void LANGUAGE plpgsql SECURITY DEFINER SET search_path = public AS $$ BEGIN END; $$;
```

The real, rate limited implementation lives earlier in the chain in `20260531000008_fix_award_activity_xp_rate_limit.sql`. Because this stub shares the same signature, `CREATE OR REPLACE` swaps the body for `BEGIN END`. It is also the last definition of the function in the migration tree, so it wins. The filename also duplicates the earlier real migration `20260529000000_fix_xp_forgery.sql`, which suggests it was committed by mistake.

**Code Location**

- `supabase/migrations/20260611000001_fix_xp_forgery.sql`

### Steps to Reproduce

1. Apply migrations through `20260611000001`.
2. Run `select public.award_activity_xp('session_join');`.
3. Observe that `points` / leaderboard XP and `user_activity_log` are unchanged.

### Expected Behavior

`award_activity_xp` awards XP per activity (session_join 50, mentor_help 100, daily_login 20, otherwise 10), logs the activity, and enforces the daily rate limits defined in `20260531000008`.

### Actual Behavior

The function returns immediately and awards nothing.

### Impact

- Activity XP is silently dead for every user: session joins, mentor help, and daily logins all award 0.
- Gamification and leaderboard progression that depend on this RPC stop working.

### Proposed Fix

Delete `supabase/migrations/20260611000001_fix_xp_forgery.sql`. Removing it makes `20260531000008` the effective definition again, which is the correct rate limited implementation.

If a deployment already applied this stub, add a forward migration that re-creates the correct body from `20260531000008`, since migrations do not re-run automatically once applied.

### Verification

Reproduced on an isolated PostgreSQL instance with minimal stubs for the Supabase dependencies:

- After loading `20260531000008`, calling `award_activity_xp('session_join')` inserted 50 XP and one `user_activity_log` row.
- After loading `20260611000001`, a further call awarded nothing (XP unchanged).

### Suggested labels

bug, database

Anexus5919:
## Summary

Removes `supabase/migrations/20260611000001_fix_xp_forgery.sql`, a stub migration that replaced `award_activity_xp(text)` with an empty body and silently disabled all activity XP.

## Changes

- Deleted the stub migration. The effective definition reverts to `20260531000008_fix_award_activity_xp_rate_limit.sql`, the correct rate limited implementation.

## Verification

Isolated PostgreSQL run:

- With `20260531000008` loaded, `award_activity_xp('session_join')` awarded 50 XP and logged one activity row.
- After loading the stub, a further call awarded nothing.

## Related

Fixes #1633

### DIFF
--- a/supabase/migrations/20260611000001_fix_xp_forgery.sql
+++ b/supabase/migrations/20260611000001_fix_xp_forgery.sql
@@ -1,1 +1,0 @@
-CREATE OR REPLACE FUNCTION award_activity_xp(_activity_type TEXT) RETURNS void LANGUAGE plpgsql SECURITY DEFINER SET search_path = public AS $$ BEGIN END; $$;


durdana3105:
RESOLVE CI TEST

Anexus5919:
@durdana3105 Done.

PR #1788, opened 2026-07-16
fix(db): restore peer_submissions and peer_reviews tables, RLS, and policies

### LINKED ISSUE #1787: Bug: peer_submissions and peer_reviews lost their tables and RLS when migration 20260602000004 was overwritten
### Description

**Bug Summary**

The `peer_submissions` and `peer_reviews` tables no longer have a `CREATE TABLE` or `ENABLE ROW LEVEL SECURITY` migration. The file `supabase/migrations/20260602000004_peer_reviews.sql`, which originally created both tables and defined their RLS and policies, was overwritten with an unrelated `submit_peer_review` RPC migration (for #1675) that reused the same filename. As a result the RLS coverage guard (`src/test/rlsCoverage.test.ts`, from #1525) fails on main for both tables, and a fresh database has no definition for either table.

**Root Cause**

`supabase/migrations/20260602000004_peer_reviews.sql` on main now contains a `submit_peer_review` function and a header comment referring to `20260706000001_peer_review_status_rpc.sql`. Its previous content (the table definitions, `ALTER TABLE ... ENABLE ROW LEVEL SECURITY`, and the SELECT, INSERT, UPDATE, and DELETE policies for both tables) is gone, and no other migration recreates them.

**Impact**

- CI is red on main and on every open PR: `rlsCoverage.test.ts` reports that `peer_submissions` and `peer_reviews` are queried from the frontend (which embeds the public anon key) but have no `ENABLE ROW LEVEL SECURITY` migration.
- On a fresh `supabase db reset`, neither table is created, so the peer review feature and the `submit_peer_review` RPC that references them cannot work.

**Code Location**

- `supabase/migrations/20260602000004_peer_reviews.sql` (overwritten)

### Steps to Reproduce

1. Run the frontend test suite: `npx vitest run src/test/rlsCoverage.test.ts`.
2. Observe the failure listing `peer_reviews` and `peer_submissions` as missing an ENABLE ROW LEVEL SECURITY migration.

### Expected Behavior

Both tables have a migration that creates them with RLS enabled and their access policies, and the RLS coverage test passes.

### Actual Behavior

Neither table has a create or RLS migration, and the RLS coverage test fails on main.

### Proposed Fix

Add a migration that restores the `peer_submissions` and `peer_reviews` table definitions, enables RLS, and recreates their policies, written idempotently (`CREATE TABLE IF NOT EXISTS`, `ENABLE ROW LEVEL SECURITY`, and `DROP POLICY IF EXISTS` before each `CREATE POLICY`) so it is a safe no-op on databases that already have them.

### Verification

- `rlsCoverage.test.ts` passes with the migration present.
- The migration applies cleanly and is idempotent on a local PostgreSQL instance: applying it twice produces no errors, and both tables end with RLS enabled and their four policies each.

### Suggested labels

bug, database

Anexus5919:
## Summary

`peer_submissions` and `peer_reviews` lost their table definitions, RLS, and policies when `20260602000004_peer_reviews.sql` was overwritten by an unrelated RPC migration. This restores them, which also fixes the `rlsCoverage.test.ts` failure that is currently red on main and every open PR.

## Changes

- Add `20260716000000_restore_peer_review_tables_rls.sql`: recreate both tables (`CREATE TABLE IF NOT EXISTS`), enable RLS, and recreate their policies (`DROP POLICY IF EXISTS` before each `CREATE POLICY`). Fully idempotent.

## Verification

- `npx vitest run src/test/rlsCoverage.test.ts` passes with the migration present.
- Applied on a local PostgreSQL instance twice: no errors, idempotent. Both tables end with RLS enabled and four policies each.

## Related

Fixes #1787

### DIFF
--- a/supabase/migrations/20260716000000_restore_peer_review_tables_rls.sql
+++ b/supabase/migrations/20260716000000_restore_peer_review_tables_rls.sql
@@ -1,0 +1,70 @@
+-- Restore peer_submissions and peer_reviews tables, RLS, and policies dropped when 20260602000004 was overwritten by an RPC migration.
+
+CREATE TABLE IF NOT EXISTS public.peer_submissions (
+    id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
+    user_id UUID REFERENCES public.profiles(id) ON DELETE CASCADE,
+    title TEXT NOT NULL,
+    description TEXT,
+    content_url TEXT,
+    content TEXT,
+    is_anonymous BOOLEAN DEFAULT false,
+    status TEXT DEFAULT 'pending',
+    created_at TIMESTAMP WITH TIME ZONE DEFAULT timezone('utc'::text, now()) NOT NULL
+);
+
+ALTER TABLE public.peer_submissions ENABLE ROW LEVEL SECURITY;
+
+DROP POLICY IF EXISTS "Enable read access for all authenticated users" ON public.peer_submissions;
+CREATE POLICY "Enable read access for all authenticated users"
+    ON public.peer_submissions FOR SELECT
+    USING (auth.role() = 'authenticated');
+
+DROP POLICY IF EXISTS "Enable insert for authenticated users" ON public.peer_submissions;
+CREATE POLICY "Enable insert for authenticated users"
+    ON public.peer_submissions FOR INSERT
+    WITH CHECK (auth.uid() = user_id);
+
+DROP POLICY IF EXISTS "Enable update for owners" ON public.peer_submissions;
+CREATE POLICY "Enable update for owners"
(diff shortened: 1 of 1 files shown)


Anexus5919:
@durdana3105 

Heads up on why I opened this PR without waiting for assignment: the CI test job is currently failing on main, and therefore on every open PR.

Since this blocks checks across the whole repo, I treated it as urgent and put up the fix right away rather than wait, so CI can go green again for everyone. The migration is idempotent (safe no-op on the existing database) and I verified rlsCoverage passes and that it applies twice cleanly on a local Postgres.

Happy to adjust anything if you would prefer a different approach. Please assign and review when you get a chance.

PR #1790, opened 2026-07-16
Add time-decayed leaderboard ranking (fixes #1699)

### LINKED ISSUE #1699: feat: Leaderboard ranking system has no decay function — a user who was active 3 months ago outranks a user who just joined and helped 20 people this week
## 🚀 Problem Statement

The README documents: *"🏆 Leaderboard & Ranking System."* The Supabase schema in `supabase/` stores user scores/points that feed the leaderboard. However, if the ranking is based purely on cumulative score, a user who was highly active 3 months ago and has since become inactive will permanently outrank a newly active user — regardless of recent contributions.

This kills motivation for new users: *"Why should I help people if I'll never reach the top because someone earned 5,000 points last semester?"*

## Proposed Fix

Add a time-decay function to the leaderboard score calculation using a Supabase database function:

```sql
-- supabase/migrations/ — add time-decayed leaderboard score

CREATE OR REPLACE FUNCTION get_decayed_leaderboard(
  decay_days INTEGER DEFAULT 30,
  limit_count INTEGER DEFAULT 50
)
RETURNS TABLE(
  user_id UUID,
  username TEXT,
  decayed_score NUMERIC,
  raw_score NUMERIC,
  rank BIGINT
) AS $$
BEGIN
  RETURN QUERY
  SELECT
    u.id AS user_id,
    u.username,
    -- Time-decay: score * e^(-lambda * days_since_activity)
    -- lambda = 0.023 → half-life of 30 days (score halves every month)
    SUM(
      a.points * EXP(-0.023 * EXTRACT(EPOCH FROM (NOW() - a.created_at)) / 86400)
    ) AS decayed_score,
    SUM(a.points) AS raw_score,
    RANK() OVER (ORDER BY SUM(
      a.points * EXP(-0.023 * EXTRACT(EPOCH FROM (NOW() - a.created_at)) / 86400)
    ) DESC) AS rank
  FROM user_activities a
  JOIN profiles u ON u.id = a.user_id
  WHERE a.created_at > NOW() - INTERVAL '1 year'  -- Cap history at 1 year
  GROUP BY u.id, u.username
  ORDER BY decayed_score DESC
  LIMIT limit_count;
END;
$$ LANGUAGE plpgsql;
```

In the frontend leaderboard component, switch from a static ranking query to this RPC call:

```typescript
const { data: leaderboard } = await supabase.rpc('get_decayed_leaderboard', {
  decay_days: 30,
  limit_count: 50,
});
```

Show both `decayed_score` (what determines rank) and `raw_score` (total all-time contribution) in the leaderboard table so users understand the system.

## Files to Create/Modify

| File | Change |
|---|---|
| `supabase/migrations/` | New migration adding `get_decayed_leaderboard` SQL function |
| `src/pages/` (leaderboard page) | Call RPC instead of direct table query |
| `src/components/` (leaderboard table) | Display decayed score and raw score |

**Suggested labels:** `enhancement`, `backend`, `frontend`, `level: intermediate`

I would like to work on this. Could you please assign it to me?

aaniya22:
## Summary
Adds time-decayed scoring to the leaderboard so recently active users aren't permanently outranked by users who accumulated XP long ago and have since gone inactive.

Closes #1699

## Problem
The leaderboard ranked users purely by cumulative XP. A user who was highly active months ago but has since stopped contributing would permanently outrank a newly active user, even if the new user is currently helping far more people. This discourages new contributors from engaging.

## Changes

**`supabase/migrations/20260716000001_add_leaderboard_time_decay.sql`**
- New `get_decayed_leaderboard(decay_days, limit_count, p_filter)` RPC function
- Applies exponential time decay to each user's XP based on `updated_at`, using a configurable half-life (default 30 days)
- Reuses the existing `Weekly` / `Monthly` / `All Time` filter cutoff convention from `get_user_rank_with_filter`
- Returns both `decayed_score` (drives ranking) and `raw_score` (unmodified XP) so the UI can show both
- Follows existing repo convention: `SECURITY DEFINER SET search_path = public`

**`src/pages/Leaderboard.tsx`**
- Swaps the direct `.from("leaderboard").order("xp")` query for the new `get_decayed_leaderboard` RPC
- Leaderboard rows and the podium now display the decayed score, with raw XP shown alongside it
- Badge lookup (`getBadgeByXP`) updated to use `raw_score` instead of the now-absent `xp` field on RPC rows
- User's own stats card (`myEntry`) still reads directly from the `leaderboard` table and is unaffected — only the ranked list uses decay

## Notes for reviewer
The issue's proposed SQL referenced a `user_activities` + `profiles` join, but this repo's schema stores XP directly on the `leaderboard` table (confirmed no `user_activities` table exists). I adapted the decay logic to work against the actual schema instead, keeping it consistent with the existing `get_user_rank` function's structure.

Happy to adjust the default decay half-life (currently 30 days) if a different value is preferred.

## Summary by CodeRabbit

* **New Features**
  * Leaderboard scores now use time-decayed scoring to reflect recent activity.
  * Weekly, monthly, and all-time leaderboard views are supported with updated rankings.
  * Scores display the decayed value, with raw scores shown when available.
  * Podium rankings now use the updated scoring system while retaining XP context.

### DIFF
--- a/src/pages/Leaderboard.tsx
+++ b/src/pages/Leaderboard.tsx
@@ -92,10 +92,10 @@ const LeaderboardRow = memo(({ entry, index, isCurrentUser }: LeaderboardRowProp
           <Flame className="h-5 w-5 text-yellow-400" />
           <div>
             <p className="text-xs text-gray-400">
-              Score {entry.raw_score !== undefined && `(${entry.raw_score} raw)`}
+              Score {entry.raw_score != null && `(${entry.raw_score} raw)`}
             </p>
             <h2 className="text-xl font-black text-cyan-400">
-              {entry.decayed_score !== undefined ? Math.round(entry.decayed_score) : entry.xp}
+              {entry.decayed_score != null ? Math.round(entry.decayed_score) : entry.xp}
             </h2>
           </div>
         </div>
@@ -502,7 +502,7 @@ const Leaderboard = () => {
                   </h3>
 
                   <p className="text-sm text-gray-400">
-                    {entry.decayed_score !== undefined ? Math.round(entry.decayed_score) : entry.xp} XP
+                    {entry.decayed_score != null ? Math.round(entry.decayed_score) : entry.xp} XP
                   </p>
 
                   <div
--- a/src/pages/Leaderboard.tsx
+++ b/src/pages/Leaderboard.tsx
@@ -31,6 +31,8 @@ interface LeaderboardEntry {
   sessions_joined: number;
   badges: string[];
   updated_at?: string;
+  decayed_score?: number;
+  raw_score?: number;
 }
 
 type LeaderboardRowProps = {
@@ -89,8 +91,12 @@ const LeaderboardRow = memo(({ entry, index, isCurrentUser }: LeaderboardRowProp
         <div className="flex items-center gap-2 rounded-2xl bg-cyan-500/10 px-4 py-3">
           <Flame className="h-5 w-5 text-yellow-400" />
           <div>
-            <p className="text-xs text-gray-400">XP</p>
-            <h2 className="text-xl font-black text-cyan-400">{entry.xp}</h2>
+            <p className="text-xs text-gray-400">
+              Score {entry.raw_score !== undefined && `(${entry.raw_score} raw)`}
+            </p>
+            <h2 className="text-xl font-black text-cyan-400">
+              {entry.decayed_score !== undefined ? Math.round(entry.decayed_score) : entry.xp}
+            </h2>
           </div>
         </div>
       </div>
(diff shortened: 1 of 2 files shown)


aaniya22:
The `Vercel` check failure is expected — it requires deploy authorization that fork PRs don't have, unrelated to this change.

For `CI / test`: all 174 tests across 23 test files passed. The job failure comes from an unhandled exception during `backend/tests/uploadPhoto.test.js`:

Error: ENOENT: no such file or directory, open '/tmp/19054bfc0738f1543227c888a23e5031'

This happens after the "ignores an attempted path-traversal filePath" test and looks like a temp-file cleanup race condition in that test file — unrelated to the leaderboard/decay changes in this PR (no upload/file-related code was touched). Happy to investigate further if it turns out to be caused by this branch, but wanted to flag it in case it's a known flaky test on main. Let me know if a re-run would help confirm.

PR #1791, opened 2026-07-16
fix(ai): reject mock interview role injection

### LINKED ISSUE #1728: Security: Mock Interview AI endpoint allows arbitrary role injection and prompt overriding
### Exact File Paths
`backend/controllers/aiController.js` (Lines 319 - 328)

### The Core Problem
The `conductMockInterview` endpoint takes a `messages` array from the client and maps it directly into `openRouterMessages` via `m.role || "user"`. Unlike the `askAI` endpoint (which strictly validates that `m.role` is only `"user"` or `"assistant"`), this endpoint performs zero validation on the roles of the incoming message history.

### The Reproducible Scenario / Impact
A malicious user can intercept the API request and inject previous messages with `"role": "system"`. This completely overwrites the AI's internal prompt/persona. An attacker can easily jailbreak the AI to bypass the "mock interview" constraints, effectively utilizing the project's OpenRouter API key as a free, unrestricted proxy for LLM completion.

### Step-by-Step Suggested Solution
1. Add a validation loop in `conductMockInterview` before processing the messages array.
2. Iterate through `messages` and enforce `if (m.role !== "user" && m.role !== "assistant") throw new HttpError(400, "Invalid role");`.
3. Filter out or reject any attempts to pass `system` roles from the client payload.

nyxsky404:
Fixes #1728

Rejects mock interview history entries unless their role is user or assistant. This prevents client-supplied system messages from being forwarded to OpenRouter and adds a controller-level regression test for the bypass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mock interview request validation by enforcing that each message is a valid object with a supported role.
  * Added clearer client error responses for unsupported or malformed message entries, preventing invalid requests from reaching the AI service.
* **Tests**
  * Added unit coverage to confirm invalid message roles (including `system`) and malformed message entries return a 400 error with the expected message.
  * Verified that external AI processing is not triggered for invalid inputs.


durdana3105:

durdana3105:

nyxsky404:
@durdana3105 done

PR #1907, opened 2026-08-03
Fix: Enforce server-side-only leaderboard score updates

### LINKED ISSUE #1889: Leaderboard points updated directly from the client without server-side verification - any user can inflate their own score
## Problem

The leaderboard ranking system allows clients to update point values via
direct Supabase client calls without any server-side verification of the
legitimacy of the point change. A user who opens the browser console can call:

```javascript
supabase
  .from('leaderboard')
  .update({ points: 99999 })
  .eq('user_id', supabase.auth.getUser().then(u => u.data.user.id))
```

Because RLS policies (if present) only check that the authenticated user is
updating their own row, but do not verify the magnitude or source of the
point change, any user can set their score to any value.

## Impact

- Leaderboard is gameable by anyone with minimal technical knowledge.
- The feature loses all value as a motivation mechanism.
- Top-ranked positions can be claimed fraudulently.

## Suggested Fix

1. Never allow clients to set absolute point values. Instead, award points
   only through server-side logic (Supabase Edge Functions or database
   triggers) triggered by verifiable events (e.g. session completion,
   question answered).
2. Create a Supabase database trigger on the `leaderboard` table that
   validates point increments come from an approved event source.
3. Add an RLS policy that blocks any direct `UPDATE` to the `points` column
   from the client:
   ```sql
   CREATE POLICY "no_client_point_update" ON leaderboard
     FOR UPDATE USING (false);
   ```
   Allow updates only via the service role key in an Edge Function.

anshul23102:
## Summary

Prevents leaderboard score manipulation by enforcing server-side-only score updates and adding comprehensive security documentation.

## Problem

Users can directly update their leaderboard points via browser console:
```javascript
supabase.from('leaderboard').update({ xp: 99999 }).eq('user_id', userId)
```

Without explicit enforcement, even with RLS policies in place, there are potential attack vectors through direct table updates.

## Solution

- Add explicit DENY policy to block all direct client updates
- Create restrictive UPDATE policy with strict WITH CHECK conditions
- Ensure score columns (xp, streak, badges) are immutable via direct updates
- Only profile fields (username, avatar_url) can be updated by users
- Comprehensive documentation of the security model

### RLS Policies
- `deny_client_score_updates`: Explicit DENY on all UPDATE attempts
- `users_can_update_profile_only`: Allows ONLY profile updates, rejects score changes
- `users_can_read_leaderboard`: Public read access for transparency

### Immutable Score Columns
Via WITH CHECK conditions:
- `xp = OLD.xp` - Cannot be modified
- `streak = OLD.streak` - Cannot be modified
- `sessions_joined = OLD.sessions_joined` - Cannot be modified
- `badges = OLD.badges` - Cannot be modified

### Server-Side Score Awards
Points only awarded through:
1. `award_activity_xp(_activity_type, _reference_id)` RPC
   - Rate-limited per user
   - Activity type validated against whitelist
   - Idempotent via reference_id

2. `increment_xp(user_id, xp_amount)` function
   - Uses SECURITY DEFINER (runs as role owner)
   - Atomic updates prevent race conditions

3. `award_badge(target_user_id, badge_name)` function
   - Prevents duplicate badges
   - Validates badge names

## Testing

- ✓ Syntax verified
- ✓ Follows existing migration patterns
- ✓ Compatible with existing security model
- ✓ Integrates with audit logging (leaderboard_updates table)

## Security Checklist

✅ Client cannot directly modify xp
✅ Client cannot directly modify streak  
✅ Client cannot directly modify badges
✅ Client cannot directly modify sessions_joined
✅ Only profile fields updateable (username, avatar)
✅ Server-side RPCs validate all score awards
✅ Audit trail logs all modification attempts
✅ RLS policies prevent privilege escalation
✅ Complete documentation of security model

## Related Issues

Fixes #1889

### DIFF
--- a/src/components/Chatbot/ChatMessage.tsx
+++ b/src/components/Chatbot/ChatMessage.tsx
@@ -1,30 +1,25 @@
-import React from "react";
+import React, { Suspense } from "react";
 import { Message } from "@/hooks/useChatbot";
+import { MarkdownRenderer } from "@/components/MarkdownRenderer";
 
 interface ChatMessageProps {
   message: Message;
 }
 
 export const ChatMessage = React.memo(function ChatMessage({ message }: ChatMessageProps) {
-  // 💻 Code formatting (fixed)
-  const formatMessage = (text: string) => {
-    if (text.includes("```")) {
-      const parts = text.split("```");
-
-      return parts.map((part, i) =>
-        i % 2 === 1 ? (
-          <pre
-            key={i}
-            className="bg-black text-green-400 p-2 rounded text-xs overflow-x-auto"
-          >
-            {part}
-          </pre>
-        ) : (
-          <span key={i}>{part}</span>
-        )
-      );
+  const renderMessageContent = (text: string) => {
+    if (message.role === "user") {
+      return <span className="whitespace-pre-wrap">{text}</span>;
     }
-    return <span>{text}</span>;
+
+    return (
+      <Suspense fallback={<span className="whitespace-pre-wrap">{text}</span>}>
+        <MarkdownRenderer
+          content={text}
+          className="prose-sm prose-invert text-sm whitespace-pre-wrap"
+        />
+      </Suspense>
+    );
   };
 
   return (
@@ -35,7 +30,7 @@ export const ChatMessage = React.memo(function ChatMessage({ message }: ChatMess
           : "bg-gray-800"
       }`}
     >
-      {formatMessage(message.text)}
+      {renderMessageContent(message.text)}
     </div>
   );
 });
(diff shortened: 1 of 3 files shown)


anshul23102:
Thank you for reviewing! I've implemented comprehensive enforcement of server-side-only leaderboard score updates for Issue #1889.

The fix includes:
- Explicit DENY policies blocking direct client updates
- Restrictive WITH CHECK conditions making scores immutable
- Server-side RPCs for all score awards with rate limiting
- Comprehensive security documentation and attack prevention model
- Integration with existing audit logging system

This prevents users from manipulating their scores through any attack vector, including browser console or direct API calls.

If the implementation looks good, would you kindly apply the **gssoc-approved** label? I'd also recommend adding **security** and **database** labels to track this security fix.

Looking forward to your feedback!

